PR #4, opened 2016-05-24
add ios9 Contacts framework check

joenoon:
I don't know if this change would fallback gracefully or not on ios < 9, so I'm not sure if you want this PR, but since I needed this as I am using the Contacts framework I figured I'd submit a PR.


yonahforst:
Thanks @joenoon !

PR #14, opened 2016-07-21
support pre-ES7 (includes is undefined)

king6cong:


yonahforst:
I don't think this will work. `in` takes an index number when used on an Array (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in).
If you want to support pre-ES7 try using indexOf

king6cong:
Thanks. Fix commited.

yonahforst:
thanks!

PR #17, opened 2016-07-27
More iOS 8 fixes

darinwilson:
This adds a couple of changes which restore functionality to iOS 8.

1) Reverse the changes made in #15, and set `IPHONEOS_DEPLOYMENT_TARGET` back to `8.0`

According to [Apple's documentation:](https://developer.apple.com/library/ios/documentation/IDEs/Conceptual/AppDistributionGuide/ConfiguringYourApp/ConfiguringYourApp.html#//apple_ref/doc/uid/TP40012582-CH28-SW47)

> The deployment target setting specifies the lowest operating system version that your app can run on.

This sounds to me like this setting just states the minimum version that the code will run on, so setting it to `8.0` shouldn't affect iOS 9 users in any way, but will also allow the library to work on projects that target version 8 and above.

2) Remove iOS 9 conditionals in `RNPPhoto.m`

[PHPhotoLibrary was introduced in iOS 8,](https://developer.apple.com/library/ios/documentation/Photos/Reference/PHPhotoLibrary_Class/index.html) so the same code will work in iOS 8 and 9.

I've tested these changes in iOS 8 and 9, although our app only asks for Photo permissions, so I don't know for certain if this will break anything else. If there's a sample app somewhere that runs through all the permissions, I'd be happy to test it against this PR and make sure it's solid. Thanks!


yonahforst:
@darinwilson - that's a great idea! I just added an example app to the repo. I'm downloading an 8.1 simulator now to test the changes, then i'll merge

yonahforst:
tested and merged

darinwilson:
Awesome - thanks!

Arthraim:
@darinwilson nice nice nice!!

PR #25, opened 2016-08-11
Added notification support for Android

jbpin:
All is in the title ;)

### DIFF
--- a/ReactNativePermissions.js
+++ b/ReactNativePermissions.js
@@ -15,7 +15,7 @@ const RNPTypes = {
 		'reminder',
 		'bluetooth',
 		'notification',
-		'backgroundRefresh', 
+		'backgroundRefresh',
 	],
 	android: [
 		'location',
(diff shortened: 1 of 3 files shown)


zifeo:
@yonahforst Any update on this :) ?

yonahforst:
@jbpin @zifeo - I'm not sure about this... this isn't a standard permission in the sense that you can only check if it's been blocked, but you can't explicitly request it. I'm open to including it if you can find a nice way to explain this in the readme :))

jasongaare:
@yonahforst I'm all for this. I was just about to do a PR myself for this very feature. 

I'm checking permissions for 'notification' on iOS to determine if I can send notifications, and it's annoying to need to write my own bridge just to check `areNotificationsEnabled` on Android.

I see you point about how you cannot 'request' this permission - but checking if the user has it 'authorized' or 'denied' would be super useful and easy to document

yonahforst:
merged!

PR #39, opened 2016-09-23
Android: Specify locale to toUpperCase call

pfeiffer:
Calls to `toUpperCase` uses device locale unless locale is specified.

This is causing some trouble on eg. Turkish devices, where `location` is upper-cased to `LOCATİON` causing a crash:

  `LOCATİON is not a constant in com.joshblour.reactnativepermissions.ReactNativePermissionsModule$RNType`

This PR solves the issue by specifying English as locale.


yonahforst:
thanks!!

Issue #45, opened 2016-10-25
[Android] Refactoring Mistake in ReactNativePermissionsModule


norman-kapschefsky:
Hi,
within [ReactNativePermissionsModule](https://github.com/yonahforst/react-native-permissions/tree/master/android/src/main/java/com/joshblour/reactnativepermissions/ReactNativePermissionsModule.java) in Line 96 

```
mPermissionsModule.requestPermission(permission, new PromiseImpl(resolve, reject));
```

you missed to replace `PromiseImpl` parameter with single`resolve` and `reject` callback. It wont build on Android in this way right now. Cheers.

yonahforst:
@norman-kapschefsky thanks for catching this. I'm busy on another project, and probably won't be able to address this until January '17. I'm happy to merge a PR, if anyone has the time to submit one!

yonahforst:
fixed in #48

PR #48, opened 2016-11-03
fixes issue #45.

robtassl:
This fixes issue #45. Which is currently preventing building to Android devices.


yonahforst:
@robtassl  - this looks great, thanks so much for submitting.

robtassl:
Hey, you might want to undo this pull. This allowed me to build in the version of RN that I was using (30), thus solving issue #45, but unfortunately it looks like it broke it for the latest version of RN (37, I just upgraded) I can verify that undoing this pull allows you to build using the latest RN version. The only real fix I can think of for this, is to add versions/releases to this project.

PR #74, opened 2017-02-07
Allow location type to be specified in getPermissionStatus.

### LINKED ISSUE #51: Logic Error in Location Permission
Hi,
from my perspective there is a problem with the location permission and their options (always, whenInUse). I my app the `location` permission with option `always` is required but if the user sets it manually to `whenInUse` via device settings the permission request responses `authorized`. I've debugged a bit and found the reason in [RNPLocation.m](https://github.com/yonahforst/react-native-permissions/blob/4a94a298e3ac66e71c33522b7c39c3c6aee6a4c7/permissions/RNPLocation.m). When doing a `request` the first thing that is done is checking the status via `getStatus`.  There, in both cases (always and whenInUse) `authorized` is returned.

```
...
        case kCLAuthorizationStatusAuthorizedAlways:
        case kCLAuthorizationStatusAuthorizedWhenInUse:
            return RNPStatusAuthorized;
...
```

Now there is no logic in `request` method that does check the option parameter (`always`) that I have passed. That's why I get `authorized` from the api even it was **not** set to `always` in settings.

Please have an eye on it.

Thanks in advance. Norman

kmagiera:
This change allows to get information whether location is authorised "whenInUse" or "always" on iOS.

Fixes #51

### DIFF
--- a/ReactNativePermissions.js
+++ b/ReactNativePermissions.js
@@ -15,7 +15,7 @@ const RNPTypes = {
 		'reminder',
 		'bluetooth',
 		'notification',
-		'backgroundRefresh', 
+		'backgroundRefresh',
 	],
 	android: [
 		'location',
@@ -57,9 +57,9 @@ class ReactNativePermissions {
 	}
 
 
-	getPermissionStatus(permission) {
+	getPermissionStatus(permission, type) {
   	if (this.getPermissionTypes().indexOf(permission) >= 0) {
-			return RNPermissions.getPermissionStatus(permission)
+			return RNPermissions.getPermissionStatus(permission, type)
 		} else {
 			return Promise.reject(`ReactNativePermissions: ${permission} is not a valid permission type on ${Platform.OS}`)
 		}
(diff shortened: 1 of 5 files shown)


yonahforst:
I wish I saw this two hours ago, before I started doing it myself 😳😂

One problem that I ran into was in `didChangeAuthorizationStatus`. I see you solved it by just passing in null. My concern is that if the user opens settings and switches from always to whenInUse, the permission will still show true until they call `getPermissionStatus('location', 'always')` again. I'm attempting to solve it by caching the requested permission type.

kmagiera:
Thanks @yonahforst for your prompt response on this!

What do you mean by "permission will still show true until they call..."?

Looks to me that the same problem applies to all the permissions, e.g. you can go to the settings and disallow camera access. The library does not currently provide API that would allow for getting notifications about permission changes (not saying that it should). We normally handle it by calling "getPermission" from app state change handler

yonahforst:
That's a really good point.

I know there's a reason we're listening to `didChangeAuthorizationStatus` i just cant remember what it was...

yonahforst:
Ok, now I remember. It's because we can't pass  completion handler when requesting permissions. the only way we know if the request was successfull or not is thru `didChangeAuthorizationStatus`.

In that case, I think your logic valid. I'm gonna undo my changes and merge your PR. Thanks!!

PR #84, opened 2017-03-14
Return Authorized in simulator

nerdmed:
In the iOS simulator the audio video permission is always granted. This makes it possible to test the code in the simulator.

### DIFF
--- a/permissions/RNPAudioVideo.m
+++ b/permissions/RNPAudioVideo.m
@@ -13,18 +13,23 @@
 @implementation RNPAudioVideo
 
 + (NSString *)getStatus:(NSString *)type
-{
-    int status = [AVCaptureDevice authorizationStatusForMediaType:[self typeFromString:type]];
-    switch (status) {
-        case AVAuthorizationStatusAuthorized:
-            return RNPStatusAuthorized;
-        case AVAuthorizationStatusDenied:
-            return RNPStatusDenied;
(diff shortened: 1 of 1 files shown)


nerdmed:
@yonahforst any feedback on this?

yonahforst:
@nerdmed - thanks so much for this PR :)

Is this to get around the denied status for microphone in the simulator?
We use `RNPAudioVideo` for camera _and_ microphone, this would kind of break camera permissions. Also, I'm not sure why a false denied status is better than a false authorized status

nerdmed:
hi @yonahforst thanks for your response actually you have microphone access inside the simulator. You just need to request it. I am not sure what happens with the camera. @ephemer @rikner do you have any feedback on this

yonahforst:
@nerdmed - interesting. So the simulator actually does have access to the microphone, but the permission status returns 'denied'? looks like its not officially supported tho: https://developer.apple.com/library/content/documentation/IDEs/Conceptual/iOS_Simulator_Guide/TestingontheiOSSimulator/TestingontheiOSSimulator.html

The status of 'denied' won't prevent this library from calling `requestAccessForMediaType:` so you can still request permission one way or the other.

I also don't know if `AVAudio​Session` behaves differently on the simulator. If it does, you can send a PR which uses `AVCaptureDevice` for camera permission and `AVAudioSession` for microphone

PR #85, opened 2017-03-20
add support for newer react-native versions and backwards compatibility for previous

alextkd:


mkoo21:
+1

yonahforst:
@alextkd - thanks so much for this PR

To be honest, I'm not so crazy about maintaining backwards compatibility for previous RN versions. This is a pretty clean fix, but RN is moving so quickly that I can easily seeing this becoming a nightmare and polluting the code. I think users should stick with older versions of this library to support older versions of RN. What do you think?

alextkd:
I would say we need to support backwards compatibility, this library will continue to add more features and as RN continues to evolve they will introduce new breaking changes and other applications will not be able to keep the step with the latest RN. If it's not bit of a deal to just put a few extra lines to support them as well, i would say to go for it because it will be very beneficial for the community and this library will become more popular.

yonahforst:
@alextkd - fair enough. merged!

Issue #83, opened 2017-03-09
upgraded to 0.27 but 'React/RCTConvert.h' file not found, breaking build


smokinjoe:
![screen shot 2017-03-09 at 10 38 45 am](https://cloud.githubusercontent.com/assets/827945/23765057/b06d0310-04b4-11e7-9fb5-f31622e89ef1.png)

I have to admit that I'm not quite sure if I did something wrong or what, but I have version 0.27 in my project, I've done the `rnpm link` and I didn't seemingly have any issues with dependencies after running `npm install`.

Would you happen to know what I done did (or forgot to do)?

If there's anything more you need please let me know.

alanhoff:
Same issue here

smokinjoe:
@alanhoff - what version of RN are you on?  I literally had the project working yesterday with RN version 0.37 and this permissions package version 0.25 and I was loving life.

I think I made some mistake last night modifying node_modules via package.json while forgetting to close the terminal bridge and now everything is completely broken.

I've tried upgrading to RN 0.40 and I only get _more_ errors that have no solutions outside of a gathering of frustrated devs here: https://github.com/facebook/react-native/issues/12042

alanhoff:
@smokinjoe locking the version to `0.2.5` worked..

`npm install --save react-native-permissions@0.2.5`

smokinjoe:
@alanhoff what version of React-Native are you running?

And are you doing any `rnpm unlink react-native-permissions` trickery before installing a different version?

I'm still getting nowhere fast.

matthijsdewit:
I fixed it taking the following actions:
http://stackoverflow.com/questions/41477241/react-native-xcode-upgrade-and-now-rctconvert-h-not-found

Besides changing the build settings in Xcode, I also made a few changes to react-native-permissions:

./RCTConvert+RNPStatus.m
```
//
//  RCTConvert+RNPermissionsStatus.m
//  ReactNativePermissions
//
//  Created by Yonah Forst on 23/03/16.
//  Copyright © 2016 Yonah Forst. All rights reserved.
//

#import "RCTConvert.h"
#import "RCTConvert+RNPStatus.h"


@implementation RCTConvert (RNPStatus)


RCT_ENUM_CONVERTER(RNPType, (@{ @"location" : @(RNPTypeLocation),
                                @"camera" : @(RNPTypeCamera),
                                @"microphone" : @(RNPTypeMicrophone),
                                @"photo" : @(RNPTypePhoto),
[...]

PR #90, opened 2017-03-30
Fix after RN 4.0 breaking changes (e.g. 'React/RCTBridgeModule.h' file not found')

traylewin:
Fixes issue: #83 .Fixing the 'Header Search Path' in the XCode project allows the compiler to find the .h files without having to append a path to any include statement. This allows us to remove the 'React/' prefixes. It also fixes another related issue...Facebook made breaking changes in react-native by moving some files under new subdirectories which broke some of the 'React/' includes.

### DIFF
--- a/RCTConvert+RNPStatus.h
+++ b/RCTConvert+RNPStatus.h
@@ -6,7 +6,7 @@
 //  Copyright © 2016 Yonah Forst. All rights reserved.
 //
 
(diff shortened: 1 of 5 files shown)


yonahforst:
@traylewin - thanks for taking the time to submit a PR <3 :)

but wouldn't this break users who are using RN > 0.40?

PR #100, opened 2017-05-18
Fix when remote notifications are registered before user settings

alexanderjarvis:
For example, when you register the device for remote notifications to
enable silent pushes `content-available: 1` and later ask the user for
the push permission but this function incorrectly returns `denied`.


sohobloo:
This makes sense.

yonahforst:
This is great. Thanks!

PR #118, opened 2017-07-07
Add support for permission request rationale

hmatthieu:
According to https://facebook.github.io/react-native/docs/permissionsandroid.html#request

We should be able to add a rationale for permission request.

Tested on Android API 23

### DIFF
--- a/index.android.js
+++ b/index.android.js
@@ -60,7 +60,7 @@ class ReactNativePermissions {
 	}
 
 
-	request(permission, rationale) {
+	request(permission, { rationale }) {
 		const androidPermission = RNPTypes[permission]
   	if (!androidPermission) return Promise.reject(`ReactNativePermissions: ${permission} is not a valid permission type on Android`);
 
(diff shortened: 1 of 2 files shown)


yonahforst:
@hmatthieu - Thanks for this pr!

`request()` already accepts a second type parameter in iOS (e.g. `request('location', 'always')`

I'm worried that using the same parameter for permission rationale on Android will make the API too cumbersome (e.g. `request('location', Platform.OS == 'ios' ? 'always' : 'we need your location')` doesn't make a lot of sense).

I'd like to support this feature so I'm open to other suggestions.

( Just off the top of my head, what if the second param was an arguments obj `{ type, rationale }` )

hmatthieu:
I updated index.android and index.ios to support an object as second argument.
Is there a similar process on iOS ? We should take advantage of this PR to support iOS rationale as well.

I didn't change the documentation. My English is not strong enough to write nice explanations.

What do you think ?

yonahforst:
@hmatthieu - this looks good. Thanks so much and sorry for the (extremely) late reply.

I'll take a stab at updating the readme tomorrow.

henrikra:
@hmatthieu I am pretty sure you should update README as well to this PR 🤔

PR #120, opened 2017-07-11
Change to right API call when check multiple permissions

gyfelton:
In ReadMe.md


yonahforst:
Yes! merged. thanks @gyfelton

PR #123, opened 2017-07-24
Add READ_SMS and RECEIVE_SMS permissions

### LINKED ISSUE #107: Sms permission
Hi! Is there a way to ask for SMS_RECEIVE permission?

vspedr:
Closes #107 

To be honest I'm not sure of what's the difference between `READ` and `RECEIVE`  in this case, because when I tested with the example app, granting one of those automatically granted the other as well. I'd guess that `RECEIVE` is for listening to incoming SMS and `READ` for reading from existing SMS. I bet it's probably safe to break it down to a single `sms` case, so if you're willing to change this just let me know :D

Edit: Tested on 6.0.0 emulator and 6.0.1 physical device, both working


yonahforst:
merged! thanks @vspedr

PR #131, opened 2017-08-16
Switch from using rnpm to react-native cli in setup guide

cibelius:
* react-native CLI is the modern way to go and should be encouraged. (rnpm is deprecated)
* also fix minor typo


yonahforst:
merged! thanks @Keksike!!

PR #136, opened 2017-09-02
fix checkMultiple handling of permissions with type on iOS

erezrokah:
Related to this:
https://github.com/yonahforst/react-native-permissions/issues/133

This is just a bug fix.
If we want to support passing 'types' to `checkMultiple` I think a different pr is required (and discussed since it can be implemented a few ways).

### DIFF
--- a/index.ios.js
+++ b/index.ios.js
@@ -39,6 +39,8 @@ class ReactNativePermissions {
   	if (!RNPTypes.includes(permission)) {
 			return Promise.reject(`ReactNativePermissions: ${permission} is not a valid permission type on iOS`);
 		}
+				
+		type = type || DEFAULTS[permission]
 		
 		return RNPermissions.getPermissionStatus(permission, type);
 	}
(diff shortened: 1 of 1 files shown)


eosterberg:
this skould be merged immediately, bumped into the same bug and ended up with the exact same fix!

th317erd:
Please merge this PR... I have also run into this issue.

index.ios.js
```diff
@@ -39,6 +39,8 @@ class ReactNativePermissions {
   	if (!RNPTypes.includes(permission)) {
 			return Promise.reject(`ReactNativePermissions: ${permission} is not a valid permission type on iOS`);
 		}
+				
+		type = type || DEFAULTS[permission]
```
yonahforst: I don't think this line is necessary. The default is handled in the objc code.

yonahforst (changes requested):

yonahforst:
@erezrokah - thanks so much for this PR. And great spot with the index bug! have a look at my comment on line 43. If you remove that line, I'll merge this ASAP

erezrokah:
@yonahforst done :) Thanks!

yonahforst (approved):

yonahforst:
thanks @erezrokah!!

PR #163, opened 2017-09-28
Fixed request logic for OS versions prior to Android M (6.0 that is)

bienur:


yonahforst:
merged! thanks @bienur

PR #184, opened 2017-11-19
Fix RN 0.48+ warning about requiresMainQueueSetup

oblador:
Fixes the warning

```
Module ReactNativePermissions requires main queue setup since it overrides `init` but doesn't implement `requiresMainQueueSetup. In a future release React Native will default to initializing all native modules on a background thread unless explicitly opted-out of.
```


zoontek:
@oblador Thanks!

PR #187, opened 2017-11-21
Update Readme with Cocoapods installation info

jhildensperger:
Feel free to merge and/or edit as you wish, but it would be great to have the Cocoapods installation instructions for those who might need it.


zoontek:
Thanks! It's a great idea indeed

PR #195, opened 2017-12-06
Add `NSPhotoLibraryAddUsageDescription` for Info.plist doc

MoOx:
I recently got issue in some version of iOS (apparently 10?) with this key missing.
Not sure why I needed that but as it feel redundant with the key below, but this avoid some crashes I found with sentry :)


zoontek:
Thanks! It's not a bad idea.

PR #199, opened 2017-12-10
Fixed bug with Android < 23

wassgha:
On Android < 6.0 , trying to record a video from the camera using `react-native-camera` will result in an error message `java.io.FileNotFoundException·/storage/emulated/0/...: open failed: EACCES (Permission denied)`
[This article](https://medium.com/react-native-development/fixing-problems-in-react-native-caused-by-new-permission-model-on-android-1e547f754b8) details the steps that need to be done in need to ensure the permissions are granted correctly.
### Rationale
Asking for `WRITE_EXTERNAL_STORAGE` will automatically grant `READ_EXTERNAL_STORAGE` but the opposite isn't true. Most developers need write access to storage when requesting it using `react-native-permission` so this makes more sense.


zoontek:
LGTM, thanks :)

PR #201, opened 2017-12-13
updated readme to reflect function reference

FrenchBully:
Updated readme to reflect function reference.
Resolve issue with error
`TypeError: Permissions.check is not a function`

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ export default class extends React.Component {
 
   // Check the status of a single permission
   componentDidMount() {
-    Permissions.check('photo').then(response => {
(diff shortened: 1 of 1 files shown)


zoontek:
@FrenchBully This only happen if you use commonJS imports (not the way to go in React-native)
No need to add .default if you use ESModule (`import Permissions from 'react-native-permissions'`)

zoontek:
I update the README: https://github.com/yonahforst/react-native-permissions/commit/27d863a09c796667c8a0d4bf7b529e321b36a5fa

PR #80, opened 2017-02-24
added MPMediaLibrary permissions / getStatus

hippopotamus:

PR #211, opened 2018-01-14
Ability to query mediaLibrary on ios

rory-pickering:
Some one had done a similar update #80 a while back but it wasn't maintained with the new version of this lib. 

I don't expect this to get accepted as it requires an additional framework (MediaPlayer.framework) and adding this to everyones projects for a permission that is rarely required takes it out of the original scope.

I just did a pull request anyway so if anyone does need it they can find it and use it 👍


rory-pickering:
Some one had done a similar update a while back but it wasn't maintained with the new version of this lib.

I understand this wont get accepted as adding a framework to everyones projects for a permission that is rarely required takes it out of the original scope.

I just did a pull request anyway so if anyone does need it they can use it 👍

PR #228, opened 2018-02-14
Fix of bugs related to permission location type "always/whenInUse"

### LINKED ISSUE #168: `always` gives `authorized` on `whenInUse` instead of `denied`
## Environment

XCode Version 9.0 (9A235)
iPhone X Simulator
`"react-native-permissions": "^0.2.7"`

## Steps to Reproduce

```
Permissions.requestPermission('location', 'always').then((response) => { ... });
```

screen shot 2017-10-04 at 5 06 24 pm

Press `Only While Using the App`

## Expected

The `response` in `.then((response) => { ... })` is `denied`.

## Actual

The `response` in `.then((response) => { ... })` is `authorized`

## Note

This line of code seems to show the expected intent:

https://github.com/yonahforst/react-native-permissions/blob/0ecbb90bf5bc39ee6d5a557f1429849404cc1405/permissions/RNPLocation.m#L26-L27

I suspect this bit of code is the culprit:

https://github.com/yonahforst/react-native-permissions/blob/0ecbb90bf5bc39ee6d5a557f1429849404cc1405/permissions/RNPLocation.m#L71-L74

Due to how it always passes in `getStatusForType:nil` even when the intended `type` is `'always'.

`getPermissionStatus('location', 'always')` returns `denied` as expected.

andwaal:
This PR fixes multiple bugs I encountered related to location permission and type _always_.

My initial problem was that after first requesting permission _whenInUse_ I did not manage to escalate the permission to _always_, even though this should be possible.

This was caused by the _request()_ method starting with a call to:
```javascript
NSString *status = [RNPLocation getStatusForType:nil];
```
Since we already had _kCLAuthorizationStatusAuthorizedWhenInUse_ rights from the first permission request, we would never enter:
```javascript
NSString *status = [RNPLocation getStatusForType:nil];
if (status == RNPStatusUndetermined) {
  self.completionHandler = completionHandler;
```
So the request for permission type _always_ would never be triggered.

So what I did:

* Since we never will get _RNPStatusUndetermined_ when we try to ecalate permission from _whenInUse_ to _always_, I changed to logic so that the only status that causes us not to ask for permission is _RNPStatusAuthorized_.

* Next I added a init() method to setup or _locationManger_.  According to my observations AND https://stackoverflow.com/questions/30106341/swift-locationmanager-didchangeauthorizationstatus-always-called/30107511  the delegate is called once when the _CLLocationManager_ is initialized. 
This caused all kinds of wierd behavior when trying to escalate permission, since the _CLLocationManager_ was initialized for each request, every permission request caused two _didChangeAuthorizationStatus callbacks.
In the case of trying to _escalate_ permission we would already have the permission _kCLAuthorizationStatusAuthorizedWhenInUse_ from the first request, causing us to enter: 
    ```javascript
    (void)locationManager:(CLLocationManager *)manager didChangeAuthorizationStatus: CLAuthorizationStatus)status {
      if (status != kCLAuthorizationStatusNotDetermined) {
    ````
    which then would trigger the callback to RN. This would happen just after initializing the _locationManger_ in the request method, before we actually had requested the _escalated_ permission. 
This would then return _Denied_ to the client.
By holding a variable who checks if its initial callback, we could remove the check for _kCLAuthorizationStatusNotDetermined_ in the delegate method and always treat it properly.

* Fixed #168 by keeping a copy of type in _lastTypeRequested_. By doing this we know the requested type, and may use it in _didChangeAuthorizationStatus_.

* Using the status returned in the _didChangeAuthorizationStatus_ instead of fetching it from _CLLocationManager_. This seems to remove the need for the _dispatch_after(dispatch_time)_ hack.

* Updated the example project to work with permission type _always_.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ user how the app uses this data. **Even if you don't use them**.
 So before submitting your app to the App Store, make sure that in your
 `Info.plist` you have the following keys:
 
-````xml
+```xml
 <key>NSBluetoothPeripheralUsageDescription</key>
 <string>Some description</string>
 <key>NSCalendarsUsageDescription</key>
@@ -244,6 +244,7 @@ So before submitting your app to the App Store, make sure that in your
 <key>NSMotionUsageDescription</key>
 <string>Some description</string>
 ```
+
 This is required because during the phase of processing in the App Store
 submission, the system detects that you app contains code to request the
 permission `X` but don't have the `UsageDescription` key and then it rejects the
@@ -279,7 +280,7 @@ Permissions.request('camera', {
 }).then(response => {
   this.setState({ cameraPermission: response })
 })
-````
+```
 
 * Permissions are automatically accepted for **targetSdkVersion < 23** but you
   can still use `check()` to check if the user has disabled them from Settings.
--- a/README.md
+++ b/README.md
@@ -60,8 +60,7 @@ react-native link react-native-permissions
    folder ➜ `Add Files to <...>`
 2. Go to `node_modules` ➜ `react-native-permissions` ➜ select
    `ReactNativePermissions.xcodeproj`
-3. Add `libReactNativePermissions.a` to `Build Phases` -> `Link Binary With
-   Libraries`
+3. Add `libReactNativePermissions.a` to `Build Phases` -> `Link Binary With Libraries`
 
 ## Using
 
@@ -139,7 +138,7 @@ Promises resolve into one of these statuses:
 | Return value   | Notes                                                                                                                                                                                                                                                                  |
 | -------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `authorized`   | User has authorized this permission                                                                                                                                                                                                                                    |
-| `denied`       | User has denied this permission at least once. On iOS this means that the user will not be prompted again. Android users can be prompted multiple times until they select 'Never ask me again'                                                                          |
+| `denied`       | User has denied this permission at least once. On iOS this means that the user will not be prompted again. Android users can be prompted multiple times until they select 'Never ask me again'                                                                         |
 | `restricted`   | **iOS** - this means user is not able to grant this permission, either because it's not supported by the device or because it has been blocked by parental controls. **Android** - this means that the user has selected 'Never ask me again' while denying permission |
 | `undetermined` | User has not yet been prompted with a permission dialog                                                                                                                                                                                                                |
 
@@ -212,8 +211,7 @@ Permissions.request('notification', { type: ['alert', 'badge'] }).then(
   will request. Open Xcode ➜ `Info.plist` ➜ Add a key (starting with "Privacy -
   ...") with your kit specific permission.
 
-Example: If you need Contacts permission you have to add the key `Privacy -
-Contacts Usage Description`.
+Example: If you need Contacts permission you have to add the key `Privacy - Contacts Usage Description`.
 
 <img width="338" alt="3cde3b44-7ffd-11e6-918b-63888e33f983" src="https://cloud.githubusercontent.com/assets/1440796/18713019/271be540-8011-11e6-87fb-c3828c172dfc.png">
 
@@ -226,7 +224,7 @@ user how the app uses this data. **Even if you don't use them**.
 So before submitting your app to the App Store, make sure that in your
 `Info.plist` you have the following keys:
 
-```xml
+````xml
 <key>NSBluetoothPeripheralUsageDescription</key>
 <string>Some description</string>
 <key>NSCalendarsUsageDescription</key>
@@ -281,7 +279,7 @@ Permissions.request('camera', {
 }).then(response => {
   this.setState({ cameraPermission: response })
 })
-```
+````
 
 * Permissions are automatically accepted for **targetSdkVersion < 23** but you
   can still use `check()` to check if the user has disabled them from Settings.
--- a/example/App.js
+++ b/example/App.js
@@ -9,6 +9,7 @@ import {
   Alert,
   AppState,
   Platform,
+  ScrollView,
 } from 'react-native'
 
 import Permissions from 'react-native-permissions'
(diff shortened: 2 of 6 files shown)


andwaal:
After more testing I found that my implementation in this PR is flawed and should probably not be merged yet.
It seems like the initial trigger of _didChangeAuthorizationStatus_ only happens when your app has never requested permission before.

This causes my implementation not to work if someone where to:
1. Start app - request _inUsePermission_
2. Terminate app
3. Open app - request _always_ rights.  -> Here my solution would case the callback never to be called.

I will try to figure out a solution for this.

andwaal:
So after working with this almost all day my conclusion is that this is _hard_.
Basicly because we got a event driven callback on the native side which we tries to fit into a single callback function.

The difficulties is more or less summarised in Apple`s documentation for _requestAlwaysAuthorization_:

>  If the current authorization status is anything other than kCLAuthorizationStatusNotDetermined, this method **does nothing and does not call the locationManager:didChangeAuthorizationStatus:** method, with **one exception.** If your app has never requested always authorization and its current authorization status is kCLAuthorizationStatusAuthorizedWhenInUse, you may **call this method one time** to try and change your app's authorization status to kCLAuthorizationStatusAuthorizedAlways.

It is this _exception_ I been struggeling with how to handle. My last attempt involved storing a boolean in NSUserDefaults to keep track of this _one_ attempt to escalate permission.
I almost got this to work as expected except for one edge case:

1. We asks user to escalate permission from _whenInUse_ to _always_
2. User select _whenInUse_(the same permission level as we previously had

The problem here is; we need to run the request to escalate permission level, but since user selected the same permission level as previously  _requestAlwaysAuthorization_ is never triggered.
So how can we then run our callback? 
Best I can think of is ugly hacks involving timeouts(if we havent gotten a _requestAlwaysAuthorization_ within x-seconds we assume user clicked _Keep when in use_) or maybe if its exists some kind of API that lets us detect the permission dialog opening/closing?

Right now I am leaning against just asking user for all the permissions at once, instead of the more elegant _ask when you need it_ approach I was hoping for.

andwaal:
I have now cleaned this and removed some non nessesary logic.
From what I could figure out the only way to detect the edge case described above is a timeout, so I added this. Right now the timeout is set to 4 seconds.

To summarise how the _escalation_ of permission from  _kCLAuthorizationStatusAuthorizedWhenInUse to kCLAuthorizationStatusAuthorizedAlways_ ended up working:


* When plugin is loaded the init method look for an entry in NSUserDefaults, if this is not found we known thtat escalated permission is not requested by this device.
* When a user request permission the first time everything is handled as usual
  * The first time we asks for permission status is always _Undetermined_
* When a user asks to escalate the permission rights we check if current permission is _kCLAuthorizationStatusAuthorizedWhenInUse_ and that the _always_ permission has not been requested before(using our entry in NSUserDefautls)
  * We now starts a timer of 4 seconds
* If user allows _always_ permission before 4 seconds is elapsed the _didChangeAuthorizationStatus_ will be triggered and we can return _RNPStatusAuthorized_ back to RN.
* If user declined _always_ permission and selects _inUseOnly_ _RNPStatusAuthorized_ will not be triggered, but our timeout will return _denied_ to the user.


The timeout is hack, but the only way I can see to actually make this work. Hopefully this is an edge case that will happen very seldom. The only backdraw I can see now is that the timeout may be to short. 
But it should also be possible to make proper precautions when calling this from RN, such as checking status after requested permission.

Hopefully this PR may be merged and be of help :)

lukasgjetting:
@andwaal 

This is great! However, I feel like it includes a lot of unneccessary logic. Please see my proposed solution in #255. Is it flawed or do you think it would work?

As it doesn't store whether we've asked for escalation before, it will be handled as if the user denied the escalation the first time (triggered by the timeout).

andwaal:
@lukasgjetting sorry about the late reply!

I have now had the time to sit down and look at your solution(using your repo https://github.com/lukasgjetting/react-native-permissions which included one additional commit after you commented above).

Your solution does indeed look lets complicated than mine(_which I must admit is a bit to verbose looking at it now_), but your solution also has a couple of flawes which I have accounted for in this PR.

- When requesting for location="always"(_when permission has not been requested before_), if the user selected "Only in use", you still return "authorized". Which is not true, since user did not grant "always", only "inUse". 
- If permission "whenInUse" has been granted, but "always" has been denied, you always return "undetermined" when requesting for permission="always". Here the plugin should return "denied".
- I also manged to crash the plugin with a "EXC_BAD_ACCESS" exception by doing the following:
- - Grant "whenInUse" permission, but deny "always" permission
- - Send multiple requests for "always" permission. 
- - ^ This will start multiple timers, which will eventualy cause some threading problems causing the exception

In my solution I store the information about the _escalation request_ between sessions using NSUserDefaults. This serves the following purpose:
- When asking for "escalated rights", we know if this has been done before. Which makes it possible to return the "denied" result right away.
- Restrict the _timeout_ walkaround to only be used the _one_ time where a user requests escalated permission, which prevents threading crashes as the one described above


Please let me know if you need more steps for reproducing these issues in our branch or if you already has discovered and fixed these on you side. So even if this PR is a bit more complicated, I do think it account for some edge cases not covered in your branch.

lukasgjetting:
@andwaal I agree - your solution seems great!

lukasgjetting:
@andwaal I just noticed what seems to be quite a huge flaw in your solution (correct me if I'm wrong!).

When checking for the permission status of location-always, it will return DENIED if the current AuthStatus is `kCLAuthorizationStatusAuthorizedWhenInUse` (meaning the user has previously granted location-when-in-use permission).

This could by fixed by checking whether an escalation has previously been asked for in the permission-check case.

````
case kCLAuthorizationStatusAuthorizedWhenInUse:
    if([type isEqualToString:@"always"]) {
        NSUserDefaults *defaults = [NSUserDefaults standardUserDefaults];
        NSString * hasEscalated = [defaults stringForKey:EscalatedServiceRequested];
        return hasEscalated == nil ? RNPStatusUndetermined : RNPStatusDenied;
    }
    return RNPStatusAuthorized;
````

Also we should wait with saving the escalation to after we request it - else it messes with the check. This is because of `didChangeAuthorizationStatus` being called twice. Once when permission is requested and once when it is granted/denied (last one doesn't happen when denying escalation, though..).

I solved this by moving it inside the dispatch (I also raised the delay to 15 sec, which works better for my use case). Though this might create problems with requesting location-always permission multiple times within the 15 second window.

````
dispatch_after(dispatch_time(DISPATCH_TIME_NOW, 15 * NSEC_PER_SEC), dispatch_get_main_queue(), ^{
    NSUserDefaults *defaults = [NSUserDefaults standardUserDefaults];
    [defaults setObject:@"YES" forKey:EscalatedServiceRequested];
    [defaults synchronize];
    self.escelatedRightsRequested = [NSNumber numberWithBool:YES];

    if(self.completionHandler){
        self.completionHandler(RNPStatusDenied);
        self.completionHandler = nil;
    }
});
````

andwaal:
Hi @lukasgjetting 
The first one was kinda intentional, as it was a fix to the bug where the plugin did return _authorized_ when asking for _always_ even if only _in use_ was granted. 
My idea was that to inform the user that it was in fact not authorized, by returning denied, was better than they way it was.
But I do see now, that we actually got enough information to return undetermined as you propose, which is a much better solution than mine!

On the second issue you raise, I am not sure if I agree. By moving saving of the escalation to after the timer ends, we will risk starting the timeout multiple times which may cause the _EXC_BAD_ACCESS_ I discussed in my previous reply. Could you elaborate what issue that occurs by setting it where it does now?

lukasgjetting:
@andwaal Hi!

When trying to escalate from whenInUse (granted) to always, `react-native-permissions` will instantly return "DENIED", before the user has a chance to accept.

This is because, as we've previously found out, `didChangeAuthorizationStatus` is called **twice**. Once before actually changing (where we want the status to be `undetermined`) and once after the user has chosen an option (where we want the status to be either `denied` or `granted`).

When we set `EscalatedServiceRequested` right before/after requesting (doesn't matter as the request is async), the status check in `didChangeAuthorizationStatus` the first time it's called will return **DENIED**, hence RNP will tell return to the JS with a "denied"-message.

Because of this, escalating is actually impossible.

I see your concern, though. I handle this myself simply by *never* requesting more than once, but I see why that doesn't really work for everyone..

I hope I managed to explain myself. If not, feel free to ask. :)

zoontek:
I am closing this since this issue does not exists anymore with the [2.0.0 release](https://github.com/react-native-community/react-native-permissions/releases/tag/2.0.0).

PR #259, opened 2018-06-22
Actually return a boolean

kevinresol:
because ios native is giving us an integer (1 or 0)


zoontek:
I am closing this since this issue does not exists anymore with the [2.0.0 release](https://github.com/react-native-community/react-native-permissions/releases/tag/2.0.0).

PR #293, opened 2019-02-12
Can not found RNPStatusRestricted in iOS

vikrambiwal:
When "location Services" off from device the we have to return the status "RNPStatusRestricted" but it returns  "RNPStatusDenied".
So I aded "location Services" check also for "RNPStatusRestricted".


zoontek:
It's now available in the `1.2.0`! Please check the changelog https://github.com/react-native-community/react-native-permissions/releases/tag/1.2.0

PR #299, opened 2019-03-22
Add e.g. ACCESS_FINE_LOCATION readme android

rtt63:
The list of android permissions has a lot of information, it can be noisy and I believe most people don't need it.

I suggest to add this line directly in README, so it could be more friendly and somebody won't need to dive in examples at all


zoontek:
It's now available in the `1.2.0`! Please check the changelog https://github.com/react-native-community/react-native-permissions/releases/tag/1.2.0

PR #317, opened 2019-08-26
Add storage options to save permission flags

h2soheili:
it is good to add option to save permissions on realm or sqlite if somebody installed one of these packages and  don't need to add async-storage for handle permissions.

i have a big project and need to reduce third packages as much as i can.

can be used to save permissions like this:

## Test Plan
 ```
import Realm from "realm";

 const ITEM_SCHEMA = {
  name: "ReduxPersistItem",
  primaryKey: "name",
  properties: {
    name: "string",
    content: "string"
  }
};

export function createRealmAccess(path = Realm.defaultPath) {
  let __realm = null;
  return async function accessRealm() {
    if (!__realm) {
      try {
        __realm = await Realm.open({
          schema: [
            ITEM_SCHEMA
          ],
          path,schemaVersion:0
        });
      } catch (error) {
        throw error;
      }
    }
    return __realm;
  };
}


export function createRealmPersistStorage() {
  const path=Realm.defaultPath
  const accessRealm = createRealmAccess(path);
  
  async function accessItemInstances() {
    const realm = await accessRealm();
    return realm.objects(ITEM_SCHEMA.name);
  }

  async function getItem(key, callback) {
    return withCallback(callback, async function() {
      const items = await accessItemInstances();
      const matches = items.filtered(`name = "${key}"`);
      if (matches.length > 0 && matches[0]) {
        return matches[0].content;
      } else {
        throw new Error(`Could not get item with key: '${key}'`);
      }
    });
  }

  async function setItem(key, value, callback) {
    return withCallback(callback, async function() {
      const realm = await accessRealm();
      realm.write(() => {
        realm.create(
          ITEM_SCHEMA.name,
          {
            name: key,
            content: value
          },
          true
        );
      });
    });
  }

  async function removeItem(key, callback) {
    return withCallback(callback, async function() {
      const realm = await accessRealm();
      const items = await accessItemInstances();
      realm.write(() => {
        const item = items.filtered(`name = "${key}"`);
        realm.delete(item);
      });
    });
  }

  return {
    getItem,
    setItem,
    removeItem
  };
}
```

now we can use this createRealmPersistStorage functions to set or get items

```
const storage = createRealmPersistStorage();
storage.setItem(ASYNC_STORAGE_KEY + permission, 'true',()=>{});
getItem(ASYNC_STORAGE_KEY + permission,(item)=>(!!item));
```


zoontek:
@club9822 Hi. Version 2.0.0 dropped the need of AsyncStorage and is currently in RC: https://github.com/react-native-community/react-native-permissions/tree/2.0.0?files=1

club9822:
> @club9822 Hi. Version 2.0.0 dropped the need of AsyncStorage and is currently in RC: https://github.com/react-native-community/react-native-permissions/tree/2.0.0?files=1

PR #322, opened 2019-09-18
Add the tvOS platform to the podspec

hufkens:
# Summary

The .podspec file was missing the tvOS platform.

## Test Plan

The change wat tested and used in a production ready app for tvOS/iOS

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| tvOS     |    ✅     |

## Checklist

### DIFF
--- a/ReactNativePermissions.podspec
+++ b/ReactNativePermissions.podspec
@@ -11,7 +11,7 @@ Pod::Spec.new do |s|
   s.authors         = package["author"]
   s.homepage        = package["homepage"]
 
-  s.platform        = :ios, "9.0"
+  s.platforms       = { :ios => "9.0", :tvos => "11.0" }
   s.requires_arc    = true
 
   s.source          = { :git => package["repository"]["url"], :tag => s.version }


mikehardy:
@hufkens - I am only commenting as an interested bystander on this repo (I use it but do not help maintain it), but since you are knowledgeable, is that style the recommended style? In react-native-device-info we have them split so it is not quite as concise and we have a different version.

https://github.com/react-native-community/react-native-device-info/blob/master/RNDeviceInfo.podspec#L15

```ruby
  s.platform     = :ios, "9.0"
  s.ios.deployment_target = '9.0'
  s.tvos.deployment_target = '10.0'
```

...your version here looks so much better.

if this one-liner is the equivalent (and more correct than) the three from device-info I can propagate this same style out to the modules I work on.

hufkens:
@mikehardy I'm not sure which one is best, but I also found this one looking nicer over having 2 entries.

mikehardy:
Thank you! - @hufkens just one final clarification - tvOS 11 vs 10? You propose 11 while device-info is on 10 and perhaps I'm behind on minimum requirements but I did not see them specified anywhere (a la the iOS 9 minimum here https://github.com/facebook/react-native#-requirements)

hufkens:
@mikehardy As far as I know all current Apple TV devices can upgrade to the latest version. So most clients will have the latest version. We can set it to 10 in stead of 11. I don't think it will matter that much.

zoontek:
@hufkens Thanks, I will make a new release on the `1.x` branch and apply the same changes on the `2.x` version

PR #323, opened 2019-09-19
Add NSBluetoothAlwaysUsageDescription for iOS 13

jlo1:
Need to add NSBluetoothAlwaysUsageDescription for iOS 13

# Summary

TestFlight builds would get rejected because the proper key wasn't added.

an iOS build that's submittable to TestFlight and pulls in react-native-permissions

Try uploading a test flight build without the keys. An email from apple will say:

```
Missing Purpose String in Info.plist - Your app's code references one or more APIs that access sensitive user data. The app's Info.plist file should contain a NSBluetoothAlwaysUsageDescription key with a user-facing purpose string explaining clearly and completely why your app needs the data.
```

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,10 @@ If you need to submit your application to the AppStore, you need to add to your 
 So before submitting your app to the App Store, make sure that in your `Info.plist` you have the following keys:
 
 ```xml
+<key>NSBluetoothAlwaysUsageDescription</key>
+<string>Some description</string>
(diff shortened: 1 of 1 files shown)


zoontek:
Thanks!

Issue #324, opened 2019-09-30
Split types out so module is easier to mock for testing


mikehardy:
# Feature Request

I'd like to see the exported types split out to a separate file, so the type-only file could be included in a mock setup file, without pulling in the logic (which checks for native module availability, and thus errors in jest test environments)

## Why it is needed

Without a separate type file you have to duplicate some of the typings just to mock things effectively, as including the main logic file causes an error related to the native module being unavailable

## Possible implementation

The netinfo module has a good example setup for this - it's mostly a file structure change:

https://github.com/react-native-community/react-native-netinfo/blob/master/src/internal/types.ts

### Code sample

Here's how you would use it https://github.com/react-native-community/react-native-netinfo/blob/master/jest.setup.js#L11

Honestly, until I tried to mock out react-native-permissions I didn't understand why the types were split out on netinfo, and then it immediately made sense

mikehardy:
In the meantime, for anyone else mocking this out in testing, I have this effectively passing my tests now (though only the 'check' method from this library is exercised in my tests - so you will need to add mocks if needed)

It is basically a copy-paste of the full type structure, then mocks on the functions - fragile, but temporarily effective

`./__mocks__/react-native-permissions.ts`:

```typescript
const ANDROID = Object.freeze({
  ACCEPT_HANDOVER: 'android.permission.ACCEPT_HANDOVER' as const,
  ACCESS_BACKGROUND_LOCATION: 'android.permission.ACCESS_BACKGROUND_LOCATION' as const,
  ACCESS_COARSE_LOCATION: 'android.permission.ACCESS_COARSE_LOCATION' as const,
  ACCESS_FINE_LOCATION: 'android.permission.ACCESS_FINE_LOCATION' as const,
  ACTIVITY_RECOGNITION: 'android.permission.ACTIVITY_RECOGNITION' as const,
  ADD_VOICEMAIL: 'com.android.voicemail.permission.ADD_VOICEMAIL' as const,
  ANSWER_PHONE_CALLS: 'android.permission.ANSWER_PHONE_CALLS' as const,
  BODY_SENSORS: 'android.permission.BODY_SENSORS' as const,
  CALL_PHONE: 'android.permission.CALL_PHONE' as const,
  CAMERA: 'android.permission.CAMERA' as const,
  GET_ACCOUNTS: 'android.permission.GET_ACCOUNTS' as const,
[...]

PR #337, opened 2019-10-06
Fix named export of constants through index (fix breaking change in 2.0.1)

draperunner:
# Summary

My app crashed after upgrading from v2.0.0 to v2.0.1 because `PERMISSIONS` and `RESULTS` were both undefined.

I am importing them directly like this:
```
import {
  check, request, PERMISSIONS, RESULTS,
} from 'react-native-permissions'
```

The bug was introduced in 067ef5d4c4bf86c0312dbce987a7488de80f3999 where these two constants are no longer exported as named exports through index.js, only through the default object.

## Test Plan

I have tested using the example app in this repo. I rewrote the App.tsx to not import the RNPermissions default, but only named exports. I verified that it crashed. I have also tested with my fix to verify that it worked.

Use react-native-permissions version 2.0.1 in an app and import `PERMISSIONS` and `RESULTS` like this: `import { PERMISSIONS, RESULTS } from 'react-native-permissions'`.

Start your app, and check the value of `PERMISSIONS` or `RESULTS`. They are undefined. Likely your app crashes instantly because it cannot access `RESULTS.GRANTED`, or other values.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -159,6 +159,8 @@ export function requestNotifications(
 
 export * from './types';
 
+export {PERMISSIONS, RESULTS} from './constants';
+
 export default {
   PERMISSIONS,
   RESULTS,


mikehardy:
I saw the same thing when I tried 2.0.1 last night


It is as a follow-on of my request to split types out for easier test consumption in #324 - which is an improvement, but I think @zoontek did not notice it was a breaking change because his example consumes the types differently https://github.com/react-native-community/react-native-permissions/blob/master/example/App.tsx#L20

I ended up reverting to 2.0.0 because I didn't have time to test the fix, but if you've made this change locally and it fixes it, that would be great - I'd love to see it released :-)

[edited to fix versions - tried 2.0.1 reverted to 2.0.0, not the other way 'round]

draperunner:
Yes, I tested it locally using the example app. I had to rewrite the imports in order to test it (not committed), since the example app uses the default import, as you mentioned.

Are you mixing 2.0.0 and 2.0.1 in your comment? I guess you tried 2.0.1 and reverted back to 2.0.0 and not the other way around? 🤔 

The workaround for this issue while waiting for a new release, is simple:

```diff
-import { check, PERMISSIONS, RESULTS } from 'react-native-permissions'
+import RNPermissions, { check } from 'react-native-permissions'
+const { PERMISSIONS, RESULTS } = RNPermissions
```

mikehardy:
Sorry for the confusion - I tried 2.0.1, encountered this, saw the difference between the example import style and my style, reverted to 2.0.0 pending further work, and you fixed it first :-)

I'll try that change + 2.0.1 via patch-package in a little bit but I'd assume the fix is good unless you hear back from me

Thanks!

zoontek (approved):

zoontek:
Oups sorry guys 🙈I did not think about named exports when splitting the files. Will do a new release immediately.

EDIT: checked it, updated the example, shipped it: https://github.com/react-native-community/react-native-permissions/releases/tag/2.0.2

mikehardy:
Confirmed 2.0.2 works for me with original code, thanks for the really quick fix @zoontek 

I still have something going on with my attempts to mock it out per #324 and I'm not good enough at typing yet to suggest a fix quickly, but I'll circle back eventually and try to nail it down

PR #338, opened 2019-10-07
NFC support

### LINKED ISSUE #288: Can NFC be supported?
Can NFC be supported?

mateosilguero:
# Summary

This Close: #288 

NFC Support

## Test Plan

Screenshot_1570411131
Simulator Screen Shot - iPhone X - 2019-10-06 at 22 19 16

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist


zoontek:
@mateosilguero Thanks! Otherwise, I made a quick check and it does not seems that NFC is a permission.

1) It is not requestable (doesn't need consent from the user).
2) It is not revokable (does not appears in app settings, neither on iOS or Android).

Screenshot 2019-10-07 at 13 44 35

It looks more like feature detection for me (a simple if to check if it's available). So it does not really make sense in this library.

mateosilguero:
I'm agree with you, i couldnt find methods to request for permission, but it tells me if NFC is or not available

mikehardy:
This appears to be the location for NFC stuff for react-native https://github.com/whitedogg13/react-native-nfc-manager - if it doesn't have the API you want they might love the contribution...

zoontek:
Agreed with @mikehardy.
A PR to export this method https://github.com/whitedogg13/react-native-nfc-manager/blob/master/ios/NfcManager.m#L7 will be more logical.

mateosilguero:
It make sense to have the isAvailable method on ios too ?

zoontek:
@mateosilguero On react-native-nfc-manager yes

PR #358, opened 2019-10-18
Fix path to AndroidManifest.xml

sonicdoe:
Fixes the path to `AndroidManifest.xml`. For example, see [the template of React Native v0.61.2](https://github.com/facebook/react-native/tree/v0.61.2/template) where the `AndroidManifest.xml` can be found in [`android/app/src/main`](https://github.com/facebook/react-native/tree/v0.61.2/template/android/app/src/main).

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ Then update your `Info.plist` with wanted permissions usage descriptions:
 
 ### Android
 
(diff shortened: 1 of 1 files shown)


zoontek:
Thanks!

PR #372, opened 2019-11-18
Dg/344 invalid permission

Dimon70007:
# Summary
Added info about ability of using this library with `use_frameworks!` tag

Sometimes we could not use modular_headers because of dependencies with precompiled dynamic frameworks. as workaround - we can manually specify dependency type for all dependencies, but dynamic frameworks
This approach will work only on cocoapods 1.7+ version
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

+ [ ] I added the documentation in `README.md`
+ [ ] I added a sample use of the API in the example project (`example/App.js`)

### DIFF
--- a/example/ios/Podfile
+++ b/example/ios/Podfile
@@ -1,4 +1,23 @@
 platform :ios, '9.0'
+
+# if you have prebuild dynamic cocoapods dependencies and could not migrate
+# to use_modular_headers you should use workaround for linking app with dynamic frameworks
+# and with static libraries by uncommenting next code
+
+# use_frameworks!
+# dynamic_frameworks = []
+# pre_install do |installer|
+#   # make all the other libraries into static frameworks by overriding the static_library
+#     installer.pod_targets.each do |pod|
+#         if !dynamic_frameworks.include?(pod.name)
+#             puts "Overriding the static_framework? method for #{pod.name}"
+#             def pod.build_type;
+#               Pod::Target::BuildType.static_library
+#             end
+#         end
+#     end
+# end
+
 require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'
 
 target 'RNPermissionsExample' do
--- a/README.md
+++ b/README.md
@@ -28,6 +28,31 @@ $ yarn add react-native-permissions
 By default no permission handler is installed. Update your `Podfile` by choosing the ones you want to check or request, then run `pod install`.
 
 ```ruby
+# if you have prebuild dynamic cocoapods dependencies and could not migrate
+# to use_modular_headers you should use workaround for linking app with dynamic frameworks
+# and with static libraries by placing this code at the top of Podfile
+# 
+# Add this code at the top of Podfile right after platform definition
+use_frameworks!
+
+dynamic_frameworks = ['RxCocoa', 'RxSwift', 'WhatEverSDKName']
+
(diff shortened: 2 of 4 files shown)


zoontek:
@Dimon70007 Hi, and thanks for the PR!

An alternative workaround could be https://github.com/react-native-community/react-native-permissions/issues/348#issuecomment-541423154. What do you think of it?

Dimon70007:
Hi @zoontek. I have read https://github.com/react-native-community/react-native-permissions/issues/348#issuecomment-541423154 and think that it is too difficult for users to manually link every static library. Even more, i have tried to use this approach (manual linking), but have no luck

example/ios/Podfile
```diff
@@ -54,3 +74,13 @@ target 'RNPermissionsExample' do
 
   use_native_modules!
 end
+
+# post_install do |installer|
```
zoontek: Id this needed in any way or can be removed?

tsconfig.json
```diff
@@ -2,10 +2,10 @@
   "compilerOptions": {
     "esModuleInterop": true,
     "jsx": "react-native",
-    "lib": ["es2015"],
+    "lib": ["es2017"],
```
zoontek: Why the changes?

RNPermissions.podspec
```diff
@@ -1,5 +1,5 @@
 require 'json'
-package = JSON.parse(File.read('./package.json'))
+package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
```
zoontek: Why changing this?

example/ios/Podfile
```diff
@@ -1,4 +1,24 @@
 platform :ios, '9.0'
+use_frameworks!
+# use_modular_headers!
+
+dynamic_frameworks = []
+
+# make all the other frameworks into static frameworks by overriding the static_framework? function to return true
```
zoontek: Not sure it's a good idea to make the example uses `use_frameworks!` be default. It could scare beginners who won't have this issue anyway.

example/ios/Podfile
Dimon70007: oh, sorry
i have forgot to remove dead code
i will do new PR soon

tsconfig.json
Dimon70007: this changes needed for compatibility with es2017 - `Object.values` syntax, used in src/index.ts

zoontek: That's not useful since Metro + react-native-babel-preset can take care of it anyway :)

example/ios/Podfile
Dimon70007: ok, but if the beginner will copy this to his podfile - it will break nothing.

RNPermissions.podspec
Dimon70007: ok. i'll drop this

example/ios/Podfile
zoontek: Of course not. But the users who don't need it don't have to think this is mandatory. Commenting the whole block + adding a similar comment as in the `README` could be a great solution.

Dimon70007: ok. done

zoontek:
Thanks!

Issue #390, opened 2019-12-31
Handler being ARC'd on iOS


booker-dragon:
# Bug
This is a problem only happens in a release app
I am experiencing issue that when I try to request for `Location When In Use` permission on iOS, the dialog pop up and close suddenly before user click the buttons

I have research that, this might be caused by iOS ARC.
https://stackoverflow.com/questions/27048995/alert-view-disappears-on-its-own-when-calling-locationmanager-requestwheninusea

I checked the source code, `locationManager` is already a strong reference in `LocationWhenInUse` class. Then I tried to make `handler` in `RNPermission.m` to a member variable  of the class instead of a local variable in `request()`. The problem solved immediately.

Though, I am not an expert of iOS, I think my fix is not a proper fix, thus I hope someone could help on fixing this.

## Environment info

React native info output:

```bash
System:
    OS: macOS 10.15.1
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 3.34 GB / 32.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.13.0 - ~/.nvm/versions/node/v12.13.0/bin/node
    npm: 6.12.0 - ~/.nvm/versions/node/v12.13.0/bin/npm
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5977832
    Xcode: 11.2.1/11B500 - /usr/bin/xcodebuild
  npmPackages:
    @react-native-community/cli: 2.9.0 => 2.9.0 
    react: 16.9.0 => 16.9.0 
    react-native: 0.61.5 => 0.61.5 
  npmGlobalPackages:
    react-native-cli: 2.0.1
```

Library version: 2.0.8

## Steps To Reproduce

1. Make your app a bit more bucky (Eating more ram)
2. Request location when in use permission on iOS
3. The dialog shown but closed itself suddenly

Describe what you expected to happen:

1. The dialog should show and wait for users to click on the option they want

## Reproducible sample code

Permission.request(PERMISSIONS.IOS.LOCATION_WHEN_IN_USE)

booker-dragon:
I have made a patch that fixed the issue but I am not sure whether it is a good fix

```
[...]

PR #391, opened 2020-01-03
Hotfix: Prevent handler being ARC on iOS before resolving the request

booker0108:
# Summary

To resolve the problem of handler being ARC on iOS [#390 ]

- Create a strong reference in RNPermission.h
- Add a method to store handler in NSDictionary and return a random ID for disposing after use

## Test Plan

- Create a memory-consuming app
- Invoke Location permission request
- Old implement might get to a situation that the dialog dismiss itself before user action because of aggressive ARC
- New implementation will not encounter the issue as handler is now stored in a strong reference to prevent ARC

- A bulky app which cause the bug (Release mode only)

- Invoke Location permission request

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌ (Not applicable)    |

## Checklist

### DIFF
--- a/ios/RNPermissions.h
+++ b/ios/RNPermissions.h
@@ -79,8 +79,6 @@ typedef enum {
 
 @interface RNPermissions : NSObject <RCTBridgeModule>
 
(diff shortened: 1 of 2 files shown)


zoontek:
LGTM 👍

PR #456, opened 2020-04-28
Fix compatibility with TypeScript versions < 3.8.0

huntie:
# Summary

In the [2.1.0 release](https://github.com/react-native-community/react-native-permissions/commit/19840ab5b0536f04c86c11cd7f733291aeb201b9#diff-41df78d86ed4de80aae9b0b9560feb2bR5-R6) of react-native-permissions, this library now exports the `Rationale` type from `react-native` in `src/types.ts`. However, the syntax used for this is a type-only import, which is a [new feature in TypeScript 3.8](https://devblogs.microsoft.com/typescript/announcing-typescript-3-8/).

https://github.com/react-native-community/react-native-permissions/blob/19840ab/src/types.ts#L5-6

This may have been done pre-emptively, as not all projects will have yet updated to a compatible TypeScript version. Since this library ships uncompiled, this will cause a syntax error when parsed by the hosting project in versions of TypeScript that are below 3.8.0, and cannot easily be ignored.

**This PR removes the type-only import and converts it back to a standard TypeScript import. There are no features impacted.**

**Before changes**

image

**After changes in a host project**

    yarn tsc -w

image

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -2,7 +2,7 @@ import {ANDROID, IOS, RESULTS} from './constants';
 
 type Values<T extends object> = T[keyof T];
 
-export type {Rationale} from 'react-native';
+export {Rationale} from 'react-native';
 
 export type AndroidPermission = Values<typeof ANDROID>;
 export type IOSPermission = Values<typeof IOS>;


zoontek:
@huntie Thanks! All our projects are already using TS 3.8 and I'm a bit used to Flow syntax.

I already forgot that it's recent in TS. 😅

mikehardy:
Understandable change but perhaps better to de-couple projects TS version from modules by proposing a PR that uses e.g. bob like the bob/build related changes here? https://github.com/react-native-community/react-native-device-info/pull/799/files

zoontek:
@mikehardy this project uses bob too: https://github.com/react-native-community/react-native-permissions/blob/master/package.json#L37

But bob uses the installed TS version to generate declaration files.

zoontek:
@huntie I made [a release](https://github.com/react-native-community/react-native-permissions/releases/tag/2.1.4) for it. Thanks again.

PR #459, opened 2020-05-04
README – Alternative to use_frameworks

joncardasis:
# Summary

Cocoapods `use_frameworks!` directive can cause problems with many RN projects, react-native-permissions included. Currently, the Readme offers a hacky workaround tied to specific versions of Cocoapods.

I've added a callout to the [cocoapods-user-defined-build-types](https://github.com/joncardasis/cocoapods-user-defined-build-types) plugin as a more stable alternative. **DISCLOSURE:** I am the maintainer of this plugin.

N/A

N/A

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ Then update your `Info.plist` with wanted permissions usage descriptions:
 ```
 
 #### Workaround for `use_frameworks!` issues
+`use_frameworks!` builds _all_ Pods as dynamic frameworks. react-native-permissions must be compiled as a static library.
 
 If you use `use_frameworks!`, add this at the top of your `Podfile`:
 
@@ -123,6 +124,8 @@ pre_install do |installer|
   end
 end
 ```
+
+Or as an alternative to `use_frameworks!`, use the [cocoapods-user-defined-build-types](https://github.com/joncardasis/cocoapods-user-defined-build-types) plugin to selectively mark other Pods as dynamic frameworks.
 
 ### Android
 


zoontek:
Hi 👋

I appreciate the idea, but a ruby for loop is not a big deal and some users have installed Cocoapods using Homebrew, not Rubygems, so it's not viable for them.

This plugin is clearly a bit more elegant, but I prefer not to confuse beginners with alternate workarounds anymore.

comvenger-brandon:
I don't understand the need for this workaround - it is the only library in our project which needs it.

Can't you just set static_framework = true in your Podspecs?

mikehardy:
Hey @zoontek I'm all the sudden very interested in the `use_frameworks` case since one of my important underlying libraries (firebase-ios-sdk) just made it a hard requirement :weary: 

Can I ask if there is a specific reason `use_frameworks` doesn't work out of the box for react-native-permissions? I'm sure I'll try the build soon enough and see myself, but if you know off-hand that would be useful. Perhaps I can try to post something that will make it build/run successfully with no Podfile work needed at all?

comvenger-brandon:
@mikehardy it does work when using use_frameworks. I think it's just a matter of updating the src here.

Firebase is also the reason why I'm interested in this one :)

mikehardy:
Okay, then sorry if I've wasted any of your time Mathieu - I'll put my head down and get back to work incrementally making progress on the whole use_frameworks thing. Most of the major components + popular repos seem to be fine, we can't be too far off :four_leaf_clover:

zoontek:
@comvenger-brandon There is a good reason for that 🙂. The core of the module (`RNPermissions.{h,m}` file) uses autolinking, like all standard react-native modules, so there is no issue with that. But in order to avoid being blocked during iOS App Store reviews (due to unused permissions requesting code in your codebase - yes, this a thing 😅), this part of the codebase is located in "handlers", which are [imported conditionally](https://github.com/zoontek/react-native-permissions/blob/master/ios/RNPermissions.m#L4).

Unfortunately, this pattern isn't currently doable when using `use_frameworks`, even by setting `s.static_framework = true` in handlers podspecs:

Screenshot 2022-08-28 at 11 47 35

Of course, I have ideas on how everything could be improved: switching to a monorepo with several packages.

- @react-native-permissions/android
- @react-native-permissions/windows
- @react-native-permissions/ios _<- only the core, autolinked, no changes required in `Podfile`_
- @react-native-permissions/ios-camera _<- handlers, **but** autolinked, no changes required in `Podfile`_
- @react-native-permissions/ios-siri
- etc…

It could improve the DX by a lot. But as usual, time, money, etc.

comvenger-brandon:
Thank you for explaining! This does make sense.

mikehardy:
I just wanted to say that I've got app store approval on my iOS app using new react-native-firebase (requires use_frameworks) and react-native-permissions (requires the workaround for use_frameworks) just this morning, and sure a podfile hack isn't the *best* DX, but hey, that workaround works. I'm quite appreciative.

PR #476, opened 2020-06-15
Update iOS setup to include pod for RNPermissions

letam:
Update README's setup for iOS: Add line to include core library in Podfile.

May save time for some.


zoontek:
You don't have to with autolinking (RN >= 0.60).

PR #510, opened 2020-09-16
Support iOS 14 photo permission API changes

jochem725:
# Summary
iOS 14 comes with some photo permission API changes that impact behavior for apps using this library.
→ Permissions are now split in “ReadWrite” and “AddOnly” permissions, so read permission is not always necessary.
→ Users can now grant “limited” permission, allowing them to pick which photos to share.

I’ve implemented both of these changes in the following way:
→ Updated the PhotoLibrary pod to make use of the new API, asking for “ReadWrite” permission (which is the old behavior)
→ Added PhotoLibraryAddOnly pod which asks AddOnly permission, (it behaves like a subset of ReadWrite, if ReadWrite is granted AddOnly is also granted)
→ When limited permission is given users can use a new method `presentLimitedLibraryPicker` to summon the limited selection picker to update the selection. This is a no-op if users give access to their full library. By making it a separate method library users are flexible when to show it (i.e. make it available via a button so users can update their shared photo subset at any time)

Impacts only iOS.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/example/ios/RNPermissionsExample/Info.plist
+++ b/example/ios/RNPermissionsExample/Info.plist
@@ -61,6 +61,8 @@
 	<string>Let me use the microphone</string>
 	<key>NSMotionUsageDescription</key>
 	<string>Let me use your motion data</string>
+	<key>NSPhotoLibraryAddUsageDescription</key>
+	<string>Let me add photos</string>
 	<key>NSPhotoLibraryUsageDescription</key>
 	<string>Let me use your photo library</string>
 	<key>NSRemindersUsageDescription</key>
--- a/ios/PhotoLibrary/RNPermissionHandlerPhotoLibrary.m
+++ b/ios/PhotoLibrary/RNPermissionHandlerPhotoLibrary.m
@@ -1,6 +1,7 @@
 #import "RNPermissionHandlerPhotoLibrary.h"
 
 @import Photos;
+@import PhotosUI;
 
 @implementation RNPermissionHandlerPhotoLibrary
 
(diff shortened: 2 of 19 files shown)


example/ios/RNPermissionsExample/Info.plist
```diff
@@ -61,6 +61,8 @@
 	<string>Let me use the microphone</string>
 	<key>NSMotionUsageDescription</key>
 	<string>Let me use your motion data</string>
+	<key>NSPhotoLibraryAddUsageDescription</key>
+	<string>Let me add photos</string>
```
AlessandroAries, line 65: also videos right?

FRizzonelli:
@zoontek Hello! We'd like to use this fix! Having this merged would be awesome!

ios/PhotoLibrary/RNPermissionHandlerPhotoLibrary.m
```diff
@@ -23,14 +31,30 @@ - (void)checkWithResolver:(void (^ _Nonnull)(RNPermissionStatus))resolve
       return resolve(RNPermissionStatusDenied);
     case PHAuthorizationStatusAuthorized:
       return resolve(RNPermissionStatusAuthorized);
+    case PHAuthorizationStatusLimited:
```
zoontek, line 34: `PHAuthorizationStatusLimited` is not available on iOS < 14, no?

FRizzonelli: I've tested it on simulator and works on both iOS 14 and iOS 13 without problems!

zoontek: The simulator might not work correctly on iOS because, unlike Android, it's not a proper emulator. Have you tested it on a real device?

zoontek: https://developer.apple.com/documentation/photokit/phauthorizationstatus/phauthorizationstatuslimited?language=occ Support is actually iOS 14+

zoontek:
@jochem725 Sorry for the delay, and thanks for that fantastic work! I merged it in the upcoming 3.0.0 version branch.

PR #519, opened 2020-09-18
fix: Xcode 12 compatibility

Fetten:
# Summary

The latest Xcode 12 fails to build when a module does not depend on `React-Core` directly. This change is necessary for all native modules on iOS. For details please see: https://github.com/facebook/react-native/issues/29633#issuecomment-694187116

## Test Plan
1. Build an App with Xcode 11.x. Result: The app builds.
2. Build an App with Xcode 12. Result: The app will not build when depending on `React`.
3. Build an App with Xcode 12 with the PR changes applied. Result: The app builds.

Mac OS with Xcode 12

see Test Plan

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/RNPermissions.podspec
+++ b/RNPermissions.podspec
@@ -3,7 +3,7 @@ package = JSON.parse(File.read('./package.json'))
 
 Pod::Spec.new do |s|
   s.name                      = "RNPermissions"
-  s.dependency                  "React"
+  s.dependency                  "React-Core"
 
   s.version                   = package["version"]
   s.license                   = package["license"]


radko93:
There are multiple podspecs in this repo so I'm not sure this is enough.

Fetten:
> There are multiple podspecs in this repo so I'm not sure this is enough.

@radko93 Yes, there are, but none depends on `React`. Instead they use the main `RNPermissions.podspec` as dependency. My tests with Xcode 12 also went all fine.

zoontek:
Thanks! I will release a new version.

geraintwhite:
Note that this change makes the minimum `react-native` version 0.60.2 https://github.com/th3rdwave/react-native-safe-area-context/pull/147

mikehardy:
I honestly have no idea how anyone can use RN60 productively any more. Doesn't this library already not support RN59?

zoontek:
@mikehardy You can use with manual link of the core + CocoaPods for the handlers + `jetifier -r` on Android.
I will update the README to drop the support of RN 0.60.

PR #538, opened 2020-11-10
Fix LimitedLibraryPicker not popping up. 

kurt-anderson-prod:
# Summary

If you have multiple ViewControllers nested, the prompt for the LimitedLibraryPicker will not prompt on the right viewcontroller, which makes nothing actually happen.

Added a quick cycler to go through the tree until it finds the top most view controller.

## Test Plan

I have run this on many different instances to make sure the code is still solid. So far I have found no issues with this small code snippet. 

None required. Just have an app with react-native-permissions installed. 

Go multiple viewcontrollers down within an app and try to call the LimitedLibraryPicker. On 3.0.0 this does not work. On the current version of the app is does. 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    -iOS Only Bug-   |

## Checklist

- [ X] I have tested this on a device and a simulator
- [ X] I added the documentation in `README.md`
- [X ] I mentioned this change in `CHANGELOG.md`
- [ X] I updated the typed files (TS and Flow)
- [ X] I added a sample use of the API in the example project (`example/App.js`)

This is a very simple commit. The above are not necassary. Not much at all has changed.

### DIFF
--- a/ios/PhotoLibrary/RNPermissionHandlerPhotoLibrary.m
+++ b/ios/PhotoLibrary/RNPermissionHandlerPhotoLibrary.m
@@ -1,4 +1,5 @@
 #import "RNPermissionHandlerPhotoLibrary.h"
+#import <React/RCTUtils.h>
 
 @import Photos;
 @import PhotosUI;
@@ -57,15 +58,7 @@
       return reject(@"cannot_open_limited_picker", @"Photo library permission isn't limited", nil);
     }
     
-    UIViewController* topViewController = [[UIApplication sharedApplication].keyWindow rootViewController];
-    UIViewController* presentedViewController = topViewController.presentedViewController;
-
-    while (presentedViewController = topViewController.presentedViewController) {
-      topViewController = presentedViewController;
-      presentedViewController = topViewController.presentedViewController;
-    }
-
-    UIViewController* rootViewController = [[UIApplication sharedApplication].keyWindow rootViewController];
+    UIViewController *topViewController = RCTPresentedViewController();
     [[PHPhotoLibrary sharedPhotoLibrary] presentLimitedLibraryPickerFromViewController:topViewController];
 
     resolve(@(true));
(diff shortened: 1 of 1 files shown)


zoontek:
Thanks! I believe this is exactly the goal of `RCTPresentedViewController` (in `RCTUtils`)

https://github.com/facebook/react-native/blob/46be292f671c70aac4ecc178c96e3a2a6a3d16da/React/Base/RCTUtils.m#L549

Could you use this instead instead ?

ios/PhotoLibrary/RNPermissionHandlerPhotoLibrary.m
```diff
@@ -56,9 +56,17 @@ - (void)openLimitedPhotoLibraryPickerWithResolver:(RCTPromiseResolveBlock _Nonnu
...
+    UIViewController* presentedViewController = topViewController.presentedViewController;
+
+    while (presentedViewController = topViewController.presentedViewController) {
+      topViewController = presentedViewController;
+      presentedViewController = topViewController.presentedViewController;
+    }
 
     UIViewController* rootViewController = [[UIApplication sharedApplication].keyWindow rootViewController];
-    [[PHPhotoLibrary sharedPhotoLibrary] presentLimitedLibraryPickerFromViewController:rootViewController];
+    [[PHPhotoLibrary sharedPhotoLibrary] presentLimitedLibraryPickerFromViewController:topViewController];
```
zoontek, line 62: Add the import:

```objc
#import <React/RCTUtils.h>
```

Replace all of this with:

```objc
UIViewController *topViewController = RCTPresentedViewController();
[[PHPhotoLibrary sharedPhotoLibrary] presentLimitedLibraryPickerFromViewController:topViewController];
```

zoontek (changes requested):

ios/PhotoLibrary/RNPermissionHandlerPhotoLibrary.m
kurt-anderson-prod: I agree, that would definitely be the better way to do it. Tested and it works!

kurt-anderson-prod:
Changes pushed up!

zoontek:
@princekurt Perfect, thank you!

zoontek (approved):

zoontek:
I released a new beta with this change 👍

PR #542, opened 2020-11-24
Fix/open settings typings

matgardon:
# Summary

* What issues does the pull request solve?
openSettings typing using a `void` return type causes an unexpected behavior/usage when consumed through an observable: 
  - an observable never emits & directly completes if consuming a promise returning `void`
  - it emits `true` then completes if consuming a promise returning `true` (boolean)

* How did you implement the solution?
fixed typings in global TS contract & ios/android/windows modules + changed the return type of the C++ windows implementation. But without much C++ knowledge and no windows machine to check ! please confirm or suggest a valid type for `boolean true` in `RNPermissions.h` & .cpp.

* What areas of the library does it impact?
ios, android and windows typings + windows implementation of `openSettings()`.

/!\ I simplified the signature of openSettings() in module.ios.ts & module.android.ts, please review & confirm if it suits the coding rules of this project.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅    |
| Windows |   ?    |

## Checklist

- [-] I have tested this on a device and a simulator
-> only partially, I don't have a windows machine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -719,7 +719,7 @@ requestMultiple([PERMISSIONS.IOS.CAMERA, PERMISSIONS.IOS.FACE_ID]).then(
 Open application settings.
 
 ```ts
-function openSettings(): Promise<void>;
+function openSettings(): Promise<true>;
 ```
 
 ```js
--- a/mock.js
+++ b/mock.js
@@ -1,7 +1,7 @@
 const {PERMISSIONS, RESULTS} = require('./lib/commonjs/constants');
 export {PERMISSIONS, RESULTS};
 
-export const openSettings = jest.fn(async () => {});
+export const openSettings = jest.fn(async () => true);
 export const check = jest.fn(async (permission) => RESULTS.GRANTED);
 export const request = jest.fn(async (permission) => RESULTS.GRANTED);
 
--- a/src/contract.ts
+++ b/src/contract.ts
@@ -7,7 +7,7 @@ import {
 } from './types';
 
 export interface Contract {
-  openSettings(): Promise<void>;
+  openSettings(): Promise<true>;
 
   check(permission: Permission): Promise<PermissionStatus>;
 
--- a/src/module.android.ts
+++ b/src/module.android.ts
@@ -34,8 +34,8 @@ function coreStatusToStatus(status: CoreStatus): PermissionStatus {
   }
 }
 
-async function openSettings(): Promise<void> {
-  await RNP.openSettings();
+function openSettings() {
+  return RNP.openSettings();
 }
 
 async function check(permission: Permission): Promise<PermissionStatus> {
--- a/src/module.ios.ts
+++ b/src/module.ios.ts
@@ -21,8 +21,8 @@ const RNP: {
   request: (permission: Permission) => Promise<PermissionStatus>;
 } = NativeModules.RNPermissions;
 
-async function openSettings(): Promise<void> {
-  await RNP.openSettings();
+function openSettings() {
+  return RNP.openSettings();
 }
 
 async function check(permission: Permission): Promise<PermissionStatus> {
--- a/src/module.ios.ts
+++ b/src/module.ios.ts
@@ -1,5 +1,5 @@
 import {NativeModules} from 'react-native';
-import {RESULTS, PERMISSIONS} from './constants';
+import {RESULTS} from './constants';
 import {Contract} from './contract';
 import {
   NotificationOption,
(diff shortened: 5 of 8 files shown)


zoontek:
@matgardon Hi. Thanks for the PR.

I'm not a fan of updating a library based on the usage of some. I personally don't use observables, and I think a lot of people don't too.

This also make the API less clear: it currently resolves or rejects if it can open the settings or not. Returning a `Promise<true>` could make people think that it will resolves with `Promise<false>` if it can't open it (I'm aware that it's not the same as `Promise<boolean>`, but some people don't use types and will don't understand the nuance)

I think wrapping it in your code could work:

```js
import { openSettings as originalOpenSettings } from "react-native-permissions";

export const openSettings = () => originalOpenSettings().then(() => true);
```

matgardon:
Hi @zoontek, thanks for taking the time to respond quickly.

I wouldn't want to update the library based on my usage, it would indeed be a bad use case for a PR.

I actually just updated the android & ios contracts to reflect their actual implementations which returns true when resolving, because a difference between a contract and it's implementation was causing a bug in my code, but I believe this is also bug prone for a number of different use cases whenever implementation and contract are unsynched.

The only implementation I changed was the windows one for consistency with the android & ios ones present in the codebase:

* android/RNPermissionsModule.java
```
  @ReactMethod
  public void openSettings(final Promise promise) {
    try {
      final ReactApplicationContext reactContext = getReactApplicationContext();
      final Intent intent = new Intent();
      final String packageName = reactContext.getPackageName();

      intent.setAction(Settings.ACTION_APPLICATION_DETAILS_SETTINGS);
      intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
      intent.setData(Uri.fromParts("package", packageName, null));

      reactContext.startActivity(intent);
      promise.resolve(true);
    } catch (Exception e) {
      promise.reject(ERROR_INVALID_ACTIVITY, e);
    }
  }
```

* ios/RNPermissions.m
```
RCT_REMAP_METHOD(openSettings,
                 openSettingsWithResolver:(RCTPromiseResolveBlock)resolve
                 rejecter:(RCTPromiseRejectBlock)reject) {
  UIApplication *sharedApplication = [UIApplication sharedApplication];
  NSURL *url = [NSURL URLWithString:UIApplicationOpenSettingsURLString];

  if (@available(iOS 10.0, *)) {
    [sharedApplication openURL:url options:@{} completionHandler:^(BOOL success) {
      if (success) {
        resolve(@(true));
      } else {
        reject(@"cannot_open_settings", @"Cannot open application settings", nil);
      }
    }];
  } else {
    [sharedApplication openURL:url];
    resolve(@(true));
  }
}
```

I already did fix the typings locally to reflect their actual implementations, but I wanted to avoid others encountering the same problem having to search for the explanation like I did. 
Let me know what you think !

zoontek:
@matgardon But the current implementations don't resolve with `true` (except of course if you avoid the JS part of the module and use `NativeModules` from RN directly):

- https://github.com/zoontek/react-native-permissions/blob/master/src/module.android.ts#L38
- https://github.com/zoontek/react-native-permissions/blob/master/src/module.ios.ts#L25

The result is not returned.

Current implementation:

Screenshot 2020-11-25 at 11 04 51

Something like this could cause issue:

Screenshot 2020-11-25 at 11 03 05

The native part return `true` because `undefined` is not yet supported by the bridge on Android & iOS and a value has to be returned. I choose `true`, it could be `null` too. None of this two make sense in this context, that's why the result is omitted in the JS side.

matgardon:
You're right, thanks for the clarification, I changed the async/await signature in the RN parts first because I thought it wasn't intended to 'forget returning', and didn't realize it made sense if you indeed wanted to return void at high level. 

That does indeed only concern my custom implementation using observables, and I actually wasn't right when saying an empty promise doesn't trigger the observable chain, this would cause a legacy compatibility issue between promises & observables. My case was that I mapped promises to their equivalent observables (in this case, `of()`) in my tests (for dark reasons with the testScheduler in observable testing with promises that doesn't handle time well when real promises are used), and the `of()` observable never emits and immediately completes, a different behavior from an empty-resolving promise.

For reference, that made me go back to promise vs. observables subtleties with this issue:https://github.com/ReactiveX/rxjs/issues/1736. Interesting read even if it's the other way around.

i'll close the PR, thanks for taking the time to respond will context :+1: and obviously for maintaining this lib, good job on the v3 !

matgardon:
as @zoontek said, this PR is not the way the `openSettings` method is intended to be used, misunderstanding on my part.

PR #551, opened 2020-12-16
Fix issue on Android 8, 9, 10 that made checkNotifications always granted

Soliman13:
# Summary

There is an issue with android that are on os version < 10, when we call checkNotifications() method, the result is always 'granted'.

Android with os version < 10

Call checkNotifications on RN, run on an android device with os version < 10, turn on the push notifs and see the result, turn off and see. 

## Compatibility

 OS    Implemented
iOS          N/A
Android   ✅

- [x] I mentioned this change in `CHANGELOG.md`

### DIFF
--- a/android/src/main/java/com/zoontek/rnpermissions/RNPermissionsModule.java
+++ b/android/src/main/java/com/zoontek/rnpermissions/RNPermissionsModule.java
@@ -3,6 +3,8 @@ package com.zoontek.rnpermissions;
 import android.Manifest;
 import android.annotation.SuppressLint;
 import android.app.Activity;
+import android.app.NotificationChannel;
+import android.app.NotificationManager;
 import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
@@ -30,6 +32,7 @@ import com.facebook.react.modules.core.PermissionAwareActivity;
 import com.facebook.react.modules.core.PermissionListener;
 
 import java.util.ArrayList;
+import java.util.List;
 
 @ReactModule(name = RNPermissionsModule.MODULE_NAME)
 public class RNPermissionsModule extends ReactContextBaseJavaModule implements PermissionListener {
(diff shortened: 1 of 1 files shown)


mikehardy:
Is it more accurate to say that this fix is just for the Android version range from `O` (8) to 10? But that it would work correctly from Marshmallow / 7 down and 10 and higher?

Soliman13:
@mikehardy Indeed thanks, the fix is from 8 included. 
It is working correctly from 10 and higher. I did not test on lower versions than 7 tho

mikehardy:
Great - thanks for replying, that was mostly intended to check that I was reading it correctly, code review works best with more than one set of eyes since we humans are so unreliable. From the perspective of fixing a "hole" of versions between 7 and 10 not inclusive, it does seem to make sense

android/src/main/java/com/zoontek/rnpermissions/RNPermissionsModule.java
```diff
@@ -145,10 +148,27 @@ private boolean permissionExists(final String permission) {
...
 
+  private boolean areNotificationsEnabled() {
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+        NotificationManager manager = (NotificationManager) getReactApplicationContext().getSystemService(Context.NOTIFICATION_SERVICE);
+        if (!manager.areNotificationsEnabled()) {
+          return false;
+        }
+        List<NotificationChannel> channels = manager.getNotificationChannels();
+        for (NotificationChannel channel : channels) {
+          if (channel.getImportance() == NotificationManager.IMPORTANCE_NONE) {
```
mikehardy, line 159: What I am curious about is how this works when:
- you have multiple channels, and a user has only made one of them silent. Notifications may still be granted just not on one channel, isn't that still okay? it does not seem to justify a "false" here

This seems to start in the direction of something semantically different then "notifications are granted" and starts being "user has all notification channels enabled" which is really a different thing, since you could just create a new notification channel I think?

Is there a reference to the actual NotificationCompat code that shows the problem requiring this workaround on these API levels? And since NotificationCompat is actually in androidx.core (which is upgradable) I would expect to see a similar google issue tracker bug showing where the problem is discovered and fixed, and then likely it would just be an androidx.core minimum version bump with documentation in the changelog explaining why?

Soliman13: You are right. Actually, I have chosen the implementation that if at least a notif channel is silenced, then the result of that method is false (as I assume it's rare to implement multiple channels on android rn).

If we want to be absolutely right we have to add some methods here to then be able to check for every channel if notifs are enabled or not. I don't think that it's a really useful case, is it?

zoontek: That's a good use case, but I don't think this is something this package should handle. It's the job of a "notifications-oriented" package imho (like https://notifee.app/)

PR #580, opened 2021-03-25
Fix homepage ref on npm page

paramaggarwal:
When I search for this library online, the first result is an npm page which links to a github repo that is different from this one. Hence this PR fixes the mismatch.


zoontek:
Good catch, thank you

PR #587, opened 2021-04-20
feat(android): add MANAGE_EXTERNAL_STORAGE permission

### LINKED ISSUE #586: Android: MANAGE_EXTERNAL_STORAGE support
# Feature request

On Android 11 (API 30), some apps need to request this permission.

## Why it is needed

Because some apps need to access external storage since Android 11.

VincentCATILLON:
# Summary

Since Android 11, the OS ignores `requestLegacyExternalStorage` property in `AndroidManifest.xml` and needs to, either remove the global external storage usage, either add another new permission added in API 30 named `MANAGE_EXTERNAL_STORAGE`.

Android recommends to not use this permission, except in some conditions like:
- File storage app
- App storing documents
- Etc

[Source](https://support.google.com/googleplay/android-developer/answer/9956427?hl=en) 

Closes https://github.com/zoontek/react-native-permissions/issues/586

## Test Plan

You only need to require it in Android 11 (API 30+).

Android 11

Request the permission `PERMISSIONS.ANDROID.MANAGE_EXTERNAL_STORAGE` and let's see if it works.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,7 @@ Add all wanted permissions to your app `android/app/src/main/AndroidManifest.xml
   <uses-permission android:name="android.permission.CALL_PHONE" />
   <uses-permission android:name="android.permission.CAMERA" />
   <uses-permission android:name="android.permission.GET_ACCOUNTS" />
+  <uses-permission android:name="android.permission.MANAGE_EXTERNAL_STORAGE" />
   <uses-permission android:name="android.permission.PROCESS_OUTGOING_CALLS" />
   <uses-permission android:name="android.permission.READ_CALENDAR" />
   <uses-permission android:name="android.permission.READ_CALL_LOG" />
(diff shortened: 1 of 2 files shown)


zoontek:
Are you sure this is requestable? The permission level isn't `dangerous` https://developer.android.com/reference/android/Manifest.permission#MANAGE_EXTERNAL_STORAGE

zoontek:
Just checked, this will not work: https://stackoverflow.com/questions/65876736/how-do-you-request-manage-external-storage-permission-in-android

VincentCATILLON:
Hi @zoontek 

Thanks for your feedback, you're right, it can't be requested like that.

Closing it.

thanhluantl2304:
any update on this? why is that not workable???

mikehardy:
It's a closed issue from months ago and the platform is the same ?
This comment seems authoritative, that is, there will be no update: https://github.com/zoontek/react-native-permissions/pull/587#issuecomment-823497555

FrederickEngelhardt:
Hello, the only thing we can do with `MANAGE_EXTERNAL_STORAGE` is detect if it's accessed according to documentation [here](https://developer.android.com/training/data-storage/manage-all-files#all-files-access).

Users can be deeplinked to the permission section on their device, but there is no direct APP way to request it. Makes sense.

Curious will a getter be added to this package? People who want to handle file storage only in RN would need something like this to alert the user they need to add the permission.

Issue #608, opened 2021-05-20
Android requestMultiple not handling blocked state properly


Iskander508:
# Bug report

## Summary

When using the `requestMultiple` and clicking the `Do not ask again` OS checkbox. The method returns state `RESULTS.DENIED` while it should be `RESULTS.BLOCKED` in some cases.

## Environment info

`react-native info` output:

```bash
System:
    OS: macOS 11.3.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 262.54 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.16.1 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 6.14.12 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.10.1 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.5, DriverKit 20.4, macOS 11.3, tvOS 14.5, watchOS 7.4
    Android SDK:
      API Levels: 28, 29, 30
      Build Tools: 28.0.3, 29.0.2, 29.0.3, 31.0.0
      System Images: android-19 | Intel x86 Atom, android-23 | Intel x86 Atom_64, android-24 | Intel x86 Atom_64, android-29 | Intel x86 Atom_64, android-29 | Google Play Intel x86 Atom, android-30 | Google APIs Intel x86 Atom
      Android NDK: Not Found
  IDEs:
    Android Studio: 4.1 AI-201.8743.12.41.7199119
    Xcode: 12.5/12E262 - /usr/bin/xcodebuild
  Languages:
    Java: 1.8.0_271 - /usr/bin/javac
    Python: 2.7.16 - /usr/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1 
    react-native: 0.63.3 => 0.63.3 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

Library version: 3.0.3

## Steps to reproduce

1. Request 2 permissions using the `requestMultiple` (for example `PERMISSIONS.ANDROID.CAMERA` and `PERMISSIONS.ANDROID.RECORD_AUDIO`)
[...]

PR #609, opened 2021-05-20
Fix android requestMultiplePermissions

Iskander508:
# Summary

Fix for https://github.com/zoontek/react-native-permissions/issues/608

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    no change     |
| Android |    ✅    |

## Checklist

- [ ] ~~I added the documentation in `README.md`~~
- [ ] I mentioned this change in `CHANGELOG.md` (where is the changelog 😄 )
- [ ] ~~I updated the typed files (TS and Flow)~~ (no change in js)
- [ ] ~~I added a sample use of the API in the example project (`example/App.js`)~~

### DIFF
--- a/android/src/main/java/com/zoontek/rnpermissions/RNPermissionsModule.java
+++ b/android/src/main/java/com/zoontek/rnpermissions/RNPermissionsModule.java
@@ -315,7 +315,6 @@ public class RNPermissionsModule extends ReactContextBaseJavaModule implements P
   public void requestMultiplePermissions(final ReadableArray permissions, final Promise promise) {
     final WritableMap output = new WritableNativeMap();
     final ArrayList<String> permissionsToCheck = new ArrayList<String>();
-    int checkedPermissionsCount = 0;
 
     Context context = getReactApplicationContext().getBaseContext();
 
@@ -324,7 +323,6 @@ public class RNPermissionsModule extends ReactContextBaseJavaModule implements P
 
       if (!permissionExists(permission)) {
         output.putString(permission, UNAVAILABLE);
-        checkedPermissionsCount++;
       } else if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
         output.putString(
           permission,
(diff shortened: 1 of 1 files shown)


mikehardy (commented):
interesting - it looks more correct / easier to read by using the necessary / already existing permissionsToCheck collection to also drive the iterations and completion condition...but, I don't understand how this fixes the issue? I am certain I'm just not reading it and seeing it, but how where was the issue in the original extra-counter style leading to the problem, such that using the collection directly fixes it? :thinking:

Iskander508:
> interesting - it looks more correct / easier to read by using the necessary / already existing permissionsToCheck collection to also drive the iterations and completion condition...but, I don't understand how this fixes the issue? I am certain I'm just not reading it and seeing it, but how where was the issue in the original extra-counter style leading to the problem, such that using the collection directly fixes it? 🤔

The "buggy" part is the last part in the `try` block. The changes before that are truly only for better readability of the code. The use of the complete `permissions` list is the mistake. When using the complete list, then later the `rationaleStatuses` array doesn't match in the callback, where only the items from `permissionsToCheck` are taken into account.

zoontek:
Could you only fix the issue and remove other changes? The android part of this module is a slightly edited version of https://github.com/facebook/react-native/blob/b161241db2ef74d2e4bff36d4972f5f0312dcc44/ReactAndroid/src/main/java/com/facebook/react/modules/permissions/PermissionsModule.java#L143 and I really want to keep it as close as the original for easier updates.

mikehardy (approved):
Now the error and correction are obvious! I think the previous version was a bit cleaner but also have a different project that has to track an upstream source so I understand now wanting better code even if it's better.

zoontek:
Just tried it, the fix works flawlessly 👍
Thanks @Iskander508

efstathiosntonas:
thanks guys, this was driving me crazy and i was using workarounds to mitigate it!

PR #628, opened 2021-07-04
fix: export PERMISSIONS

1natsu172:
# Summary

The `PERMISSIONS` that were exported in v2 are not exported in v3. Specifically, the PERMISSIONS object is exported, but the actual object is empty.

This causes an error when migrating from v2 because the permission constants are not found.

In this PR, the `PERMISSIONS` exported from each OS file has been removed and integrated into `permission.ts`.

## Test Plan

* Check the `PERMISSIONS` exported from each OS file is not used from anywhere.
* ran
	* `yarn run lint`
	* `yarn run format`
	* `yarn run tscheck`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/mock.js
+++ b/mock.js
@@ -1,6 +1,15 @@
-const {PERMISSIONS} = require('./dist/commonjs/permissions');
+const {
+  PERMISSIONS: {ANDROID},
+} = require('./dist/commonjs/permissions.android');
+const {
+  PERMISSIONS: {IOS},
+} = require('./dist/commonjs/permissions.ios');
+const {
+  PERMISSIONS: {WINDOWS},
+} = require('./dist/commonjs/permissions.windows');
 const {RESULTS} = require('./dist/commonjs/results');
 
+const PERMISSIONS = {ANDROID, IOS, WINDOWS};
 export {PERMISSIONS, RESULTS};
 
 export const openLimitedPhotoLibraryPicker = jest.fn(async () => {});


zoontek:
@1natsu172 I'm not sure to understand what this PR really fix. Exporting empty maps on other platforms is something we want to avoid bundling unnecessary permissions.

Ex: The permission list on windows weights ~8ko. There's no need to put that in the iOS or the Android JS bundle.

1natsu172:
@zoontek Sorry about that. I've been using the constants of each OS for test assertions in jest.

I meant that the automatic resolution of "platform-specific extensions" by metro is not possible in jest, so can't get the constants in the node runtime. So I tried to export it with this PR.

However, I also really understood your point about not doing unnecessary platform bundling.

What do you think of the idea of not exporting from `src/permissions.ts` but exporting the constants for each platform for testing from `mock.js`? If you agree, will modify this PR.

zoontek:
@1natsu172 I agree, only the mock needs to be edited.

```js
// mock.js
const {PERMISSIONS: {ANDROID}} = require('./dist/commonjs/permissions.android');
const {PERMISSIONS: {IOS}} = require('./dist/commonjs/permissions.ios');
const {PERMISSIONS: {WINDOWS}} = require('./dist/commonjs/permissions.windows');

const {RESULTS} = require('./dist/commonjs/results');

const PERMISSIONS = {ANDROID, IOS, WINDOWS};
export {PERMISSIONS, RESULTS};

// …
```

This should work. I don't want to export `ANDROID`, `IOS` or `WINDOWS` directly from the module, as this makes no sense / it's no clear to use in a codebase.

1natsu172:
@zoontek Thank you. I just changed the PR content to `mock.js` and re-committed. Please check it.

zoontek:
LGTM 👍

zoontek:
I made a new release for it: https://github.com/zoontek/react-native-permissions/releases/tag/3.0.5

1natsu172:
@zoontek Thanks for the very very fast release! 🚀

PR #654, opened 2021-10-07
Adding new Android 12 bluetooth permissions

jsimonassi:
# Summary

I didn't open any issue for that. I just needed to implement the new android 12 bluetooth permissions in my project and I saw in this library a simple way to solve my problem and contribute to the community. Here are the new permissions in the official Google documentation:

https://developer.android.com/about/versions/12/features/bluetooth-permissions

I implemented this solution updating the example application for android 12 (Api 31) and adding the new implementations in the necessary files (See commits).

To test, just run the demo app on an Android 12 device and request the new permissions.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅    |

## Checklist

- [X] I added a sample use of the API in the example project (`example/App.js`)

### DIFF
--- a/example/android/app/build.gradle
+++ b/example/android/app/build.gradle
@@ -196,6 +196,7 @@ dependencies {
     implementation "com.facebook.react:react-native:+"  // From node_modules
 
     implementation "androidx.swiperefreshlayout:swiperefreshlayout:1.0.0"
+    implementation('com.squareup.okhttp3:okhttp:4.9.1')
 
     debugImplementation("com.facebook.flipper:flipper:${FLIPPER_VERSION}") {
       exclude group:'com.facebook.fbjni'
--- a/README.md
+++ b/README.md
@@ -162,6 +162,9 @@ Add all wanted permissions to your app `android/app/src/main/AndroidManifest.xml
   <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
   <uses-permission android:name="android.permission.ACTIVITY_RECOGNITION" />
   <uses-permission android:name="android.permission.ANSWER_PHONE_CALLS" />
+  <uses-permission android:name="android.permission.BLUETOOTH_CONNECT" />
+  <uses-permission android:name="android.permission.BLUETOOTH_SCAN" />
+  <uses-permission android:name="android.permission.BLUETOOTH_ADVERTISE" />
   <uses-permission android:name="android.permission.BODY_SENSORS" />
   <uses-permission android:name="android.permission.CALL_PHONE" />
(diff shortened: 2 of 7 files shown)


mikehardy (commented):
Thanks for posting this! Seems fine, one thing I think worth noting is that this is a breaking change for the library. Not sure about @zoontek's policy but I release them without hesitation I just make sure the release notes are great. Just "Requires a new higher minimum compileSdkVersion of 31" and as a developer I'd know exactly how to handle it. Crucially: this is an easy change for devs, bumping compileSdkVersion is trivial.

One question - did the example still build and run? I'm just surprised it was able to do a compileSdkVersion of 31 with a buildToolsVersion of only 29.n.n? I thought buildTools needed to be same or higher but I'm not really sure on that

jsimonassi:
The latest version of BuildToolsVersion is 30.0.0 rc4 (May 2020). The changelog says the only change was "includes support for building with Android 11 visualization APIs". As it was not something that directly affected the library, I chose not to update. But that should be considered in the future, especially when something about Android 12 is mentioned. For now, the example is working fine for me. I tested it on a Pixel 5.

mikehardy:
Build Tools 31 is out:

image


However, if you have empirical evidence (that is: you tested it, it works - awesome) that buildTools may be below compileSdkVersion, that works for me

example/android/app/build.gradle
```diff
@@ -196,6 +196,7 @@ dependencies {
     implementation "com.facebook.react:react-native:+"  // From node_modules
 
     implementation "androidx.swiperefreshlayout:swiperefreshlayout:1.0.0"
+    implementation('com.squareup.okhttp3:okhttp:4.9.1')
```
zoontek, line 199: Is this mandatory? You don't use it, and `okhttp` is already a RN dependency

zoontek:
LGTM 👍
I only have a small question about `okhttp`

example/android/app/build.gradle
jsimonassi: Without updating okHttp I got this error:

Screen Shot 2021-10-07 at 08 39 42

This line updated this RN dependency and fixed my problem. Another solution would be to update React Native completely (No problem if you prefer to cancel this PR and do this).

zoontek: It's probably better to update react-native in example, yes.

jsimonassi: All good! I will do this and open a new PR in the future with the modifications if it works well. Thanks for your time reviewing this.

Issue #653, opened 2021-10-06
Support Android 12 Bluetooth Permission


brunomartinezciompi:
### Why it is needed?

Android introduced a new bluetooth permission on Android 12. 

https://developer.android.com/guide/topics/connectivity/bluetooth/permissions

React Native added support for this in version 0.66 https://reactnative.dev/blog/2021/10/01/version-066

_"The BLUETOOTH_ADVERTISE, BLUETOOTH_CONNECT, and BLUETOOTH_SCAN permissions are runtime permissions. Therefore, you must explicitly request user approval in your app before you can look for Bluetooth devices, make a device discoverable to other devices, or communicate with already-paired Bluetooth devices."_

<img width="304" alt="Screen Shot 2021-10-06 at 15 33 20" src="https://user-images.githubusercontent.com/50216253/136262557-1a6d67ee-3bd8-4e46-a024-541907def9b8.png">

This issue was closed for lack of information: https://github.com/zoontek/react-native-permissions/issues/645

But I think there should be a feature request issue open in this library just in case anyone find this issue on the repository and wants to create a pull request.

### Possible implementation

-

### Code sample

```js
-
```

mikehardy:
Here are the lists to the react-native commits, for reference:
https://github.com/facebook/react-native/commit/eeb8e5829e183f6b5cd5fd327cf6da03a7db0541
https://github.com/facebook/react-native/commit/2bcc6fac3844f0752bc7067517c92a643679575e

zoontek:
Fixed by #655

PR #655, opened 2021-10-09
Adding new Android 12 bluetooth permissions

jsimonassi:
# Summary

I needed to implement the new android 12 bluetooth permissions in my project and saw in this library a simple way to solve my problem and contribute to the community. Here are the new permissions in the official Google documentation:

https://developer.android.com/about/versions/12/features/bluetooth-permissions

I implemented this solution by updating the sample application for android 12 (Api 31) and adding the new implementations to the necessary files (see commits).

It was necessary to update the react-native version to 0.66 to fix compilation errors (An RN dependency (V. 0.64.2) was deprecated in Android API 31).

To update the react-native version, I used the Upgrade Helper as base: https://react-native-community.github.io/upgrade-helper/

This PR serves: #653 

To test, just run the demo app on an Android 12 device and request the new permissions.

I tested it on a Pixel 5 with Android 12. To validate the react-native changes, I also tested it on an iPhone SE with iOS 15.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [X] I added a sample use of the API in the example project (`example/App.js`)

### DIFF
--- a/example/android/build.gradle
+++ b/example/android/build.gradle
@@ -4,8 +4,8 @@ buildscript {
     ext {
         buildToolsVersion = "29.0.3"
         minSdkVersion = 21
-        compileSdkVersion = 29
-        targetSdkVersion = 29
+        compileSdkVersion = 31
+        targetSdkVersion = 31
         ndkVersion = "20.1.5948944"
     }
     repositories {
--- a/example/android/build.gradle
+++ b/example/android/build.gradle
@@ -2,18 +2,18 @@
 
 buildscript {
     ext {
-        buildToolsVersion = "29.0.3"
+        buildToolsVersion = "31.0.0"
         minSdkVersion = 21
         compileSdkVersion = 31
         targetSdkVersion = 31
-        ndkVersion = "20.1.5948944"
+        ndkVersion = "23.0.7599858"
     }
     repositories {
         mavenCentral()
         google()
     }
     dependencies {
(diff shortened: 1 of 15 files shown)


example/android/build.gradle
```diff
@@ -2,18 +2,18 @@
 
 buildscript {
     ext {
-        buildToolsVersion = "29.0.3"
+        buildToolsVersion = "31.0.0"
         minSdkVersion = 21
         compileSdkVersion = 31
         targetSdkVersion = 31
-        ndkVersion = "20.1.5948944"
+        ndkVersion = "23.0.7599858"
```
mikehardy, line 9: Should be a different version I think for ndkVersion ?

        ndkVersion = "21.4.7075529"

mikehardy (changes requested):
Tiny item is that the ndkVersion wasn't the same as the react-native 0.66 template

Separately - better to rebase this (I'd do `git rebase -i HEAD~3` to start an interactive rebase on last 3 commits) and drop the first commit, then re-order the 2nd and 3rd so example update was the new 1st commit and feature addition was new 3rd commit

If that's not doable for whatever reason the the example app's custom bits (the okhttp bump?) should be removed at least during the example upgrade - I don't think it's needed anymore?

zoontek:
Thanks, it works correctly on my Pixel (Android 12) too. I will made a few change and publish a release with it! 🙂

PR #656, opened 2021-10-12
Make react-native-windows peer dependency optional

geraintwhite:
In npm 7 and npm 8, `npm audit fix` will fail with `Unable to resolve dependency tree` if `react-native-windows` is not installed. As `react-native-permissions` works perfectly on mobile without `react-native-windows`, this should be an optional peer dependency.

Ref https://github.com/luggit/react-native-config/pull/585

https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,5 +79,10 @@
     "react": "16.13.1",
     "react-native": "0.63.3",
     "typescript": "^4.3.5"
+  },
+  "peerDependenciesMeta": {
+    "react-native-windows": {
+      "optional": true
+    }
   }
 }


mikehardy (approved):
This peerDependency stuff is a build breaker and getting more and more important to get just right. I think this is the right approach here, for whatever that's worth @zoontek

zoontek:
I agree with that, I will push a new version tomorrow, thanks!

zoontek:
https://github.com/zoontek/react-native-permissions/releases/tag/3.0.6

Issue #648, opened 2021-09-23
iOS 15 ATT permission not showing up


VickyStash:
### Bug summary

ATT permission prompt not showing on iOS 15
Before it was working fine for iOS 14.5 and I didn't change anything.

### Library version

3.0.3

### Environment info

```shell
SDKs:
    iOS SDK:
      Platforms: iOS 14.0, DriverKit 19.0, macOS 10.15, tvOS 14.0, watchOS 7.0
    Android SDK:
      API Levels: 23, 25, 26, 27, 28, 29, 30
      Build Tools: 27.0.3, 28.0.2, 28.0.3, 29.0.2
      System Images: android-26 | Google APIs Intel x86 Atom, android-30 | Google APIs Intel x86 Atom
      Android NDK: Not Found
  IDEs:
    Android Studio: 4.0 AI-193.6911.18.40.6626763
    Xcode: 12.0.1/12A7300 - /usr/bin/xcodebuild
  Languages:
    Java: 1.8.0_222 - /usr/bin/javac
    Python: 2.7.15 - /usr/local/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1 
    react-native: 0.63.4 => 0.63.4 
    react-native-macos: Not Found
```

zoontek:
Please respect the issue template.

VickyStash:
@zoontek I've updated the description a little
I'm not sure if it's a problem with the lib or with ios 15

zoontek:
Are you using an iPhone? Which one? The simulator?

VickyStash:
@zoontek I'm checking on iPhone 8 Plus, before update to ios 15 permission was showed as expected
Also please see https://www.reddit.com/r/iOSProgramming/comments/pt41jz/att_prompt_not_showing_on_ios_15/

VickyStash:
@zoontek any ideas?

zoontek:
Reading the Reddit thread, it seems that Apple broke the feature. The code is super basic: https://github.com/zoontek/react-native-permissions/blob/master/ios/AppTrackingTransparency/RNPermissionHandlerAppTrackingTransparency.m#L38

VickyStash:
@zoontek 
![image](https://user-images.githubusercontent.com/23176449/134535470-7bea6be3-04cf-4833-a4d1-3a7cab9c8b1c.png)

zoontek:
[...]

PR #657, opened 2021-10-14
Add ATT workaround to docs

thewirv:
# Summary
Added workaround for App Tracking Transperency on iOS 15.0 to the documentation.

Motivation: ... as discussed in [this comment](https://github.com/zoontek/react-native-permissions/issues/648#issuecomment-933461300)

## Test Plan

No testing needed, only docs have changed.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,24 @@ pre_install do |installer|
 end
 ```
 
+#### Workaround for "App Tracking Transparency not requestable on mount, starting at iOS 15.0" issue:
+
+On iOS 15.0+ requesting App Tracking Transparency does not seem to work on mount (see [#648](https://github.com/zoontek/react-native-permissions/issues/648)). We can work around this by listening for app state change, and then requesting it when the app becomes `'active'`:
+
+```js
+useEffect(() => {
+  const listener = AppState.addEventListener("change", (status) => {
+    if (Platform.OS === "ios" && status === "active") {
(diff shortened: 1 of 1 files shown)


zoontek:
You have been to fast! Check my comment: https://github.com/zoontek/react-native-permissions/issues/648#issuecomment-943397384

I suggest avoiding the wording "workaround" here since it's the correct behavior: Your app must be active to request ATT. A workaround is more of a temporary solution until the issue is properly fixed.

mikehardy:
:-) too fast is a good problem to have, but kudos for taking my not-that-gentle nudge for a PR and doing it :-) :champagne:

thewirv:
@mikehardy Sure, I mean this is how Open Source works, and I can't stand people who are just complaining, without the thought of actually helping out ever occuring to them.

@zoontek Do you think it would make sense to say something in the vein of, "ATT permissions should be requested when the App becomes active because requesting it on mount can lead to problems," or do you think the explanation is not necessary?

zoontek:
@TheWirv Instead of "on mount can lead to problems", you can say that "if the app is not active, the request prompt will not appear". Problems is a bit too vague

zoontek:
I merge it and will add a few changes since I want it to be in the README for the next release (npm one is not synced with GitHub README). Thanks a lot 🙂

thewirv:
Hey @zoontek, sorry, I spaced out there for a moment. I 100% agree with your points, but thanks for merging anyway. :)

PR #658, opened 2021-10-26
tvOS compatibility

joonmanji:
# Summary

We have a monorepo that shares podfiles across mobile and tv apps. We'd like to make certain libraries tvOS compatible

Currently affects:
- Contacts
- Notifications
- Photo Library

## Test Plan

Projects should still compile with iOS code unaffected and will now compile for tvOS.

no prereqs

does it compile?

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅     |

## Checklist

### DIFF
--- a/ios/Contacts/RNPermissionHandlerContacts.m
+++ b/ios/Contacts/RNPermissionHandlerContacts.m
@@ -1,6 +1,8 @@
 #import "RNPermissionHandlerContacts.h"
 
+#if !TARGET_OS_TV
 @import Contacts;
+#endif
 
 @implementation RNPermissionHandlerContacts
 
@@ -14,6 +16,13 @@
 
 - (void)checkWithResolver:(void (^ _Nonnull)(RNPermissionStatus))resolve
                  rejecter:(void (__unused ^ _Nonnull)(NSError * _Nonnull))reject {
+#if TARGET_OS_TV
+    NSError * error = [[NSError alloc]
+                       initWithDomain:@"RNPermissions"
+                       code:1
+                       userInfo:@{NSLocalizedDescriptionKey: @"Contacts not available on tvOS"}];
+    return reject(error);
+#elif
   switch ([CNContactStore authorizationStatusForEntityType:CNEntityTypeContacts]) {
     case CNAuthorizationStatusNotDetermined:
       return resolve(RNPermissionStatusNotDetermined);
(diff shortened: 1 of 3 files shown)


mikehardy (commented):
Not sure if you've ever hacked in tvOS support but this looks about how it's gone in other repos I've worked in @zoontek :-) - mostly just ifdef'ing it out as short circuits.

Might be nice to have a generic "make me an error with this code+message" method somewhere to reduce duplication but :shrug: 

Might be nice to just chuck all ideas of formatting at the clang-format program as a separate matter https://github.com/invertase/react-native-firebase/blob/514e6bd51e7624a6403dda706f4e5b65cee63422/package.json#L12-L13 (clang-format is an npm package that delivers the binary in usable format nicely)

all minor though, the general ifdef'ing strategy is all I've seen done normally so works for me FWIW

ios/PhotoLibrary/RNPermissionHandlerPhotoLibrary.m
```diff
@@ -16,6 +16,9 @@ + (NSString * _Nonnull)handlerUniqueId {
 
 - (void)checkWithResolver:(void (^ _Nonnull)(RNPermissionStatus))resolve
                  rejecter:(void (__unused ^ _Nonnull)(NSError * _Nonnull))reject {
+#if TARGET_OS_TV
+    return reject(@"cannot_open_limited_picker", @"Photo Library not available on tvOS", nil);
```
zoontek, line 20: Wrong message code: `cannot_check_permission`

ios/Contacts/RNPermissionHandlerContacts.m
```diff
@@ -14,6 +16,13 @@ + (NSString * _Nonnull)handlerUniqueId {
 
 - (void)checkWithResolver:(void (^ _Nonnull)(RNPermissionStatus))resolve
                  rejecter:(void (__unused ^ _Nonnull)(NSError * _Nonnull))reject {
+#if TARGET_OS_TV
```
zoontek, line 19: You mix up `NSError` based and string based rejection. Maybe you can keep it to the second?

```objc
#if TARGET_OS_TV
    NSError * error = [[NSError alloc]
                       initWithDomain:@"RNPermissions"
                       code:1
                       userInfo:@{NSLocalizedDescriptionKey: @"Contacts not available on tvOS"}];
    return reject(error);
#elif

// ⬇️

#if TARGET_OS_TV
  return reject(@"cannot_check_permission", @"Contacts not available on tvOS", nil);
#elif
```

zoontek (changes requested):

zoontek:
I'm closing this since there were no follow up.

PR #677, opened 2022-02-21
README: add explanation for RESULTS.UNAVAILABLE for location permissions

miallo:
The previous explanation was correct, but an example of how this could be provoked was missing

# Summary

We ran into an issue because for us it wasn't clear under which circumstances RESULTS.UNAVAILABLE would be returned. The existing documentation was totally correct, but we misunderstood it. Hopefully this change in the README makes it a bit easier to get an understanding of when it might be returned

## Checklist

- [x] I added documentation in `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -692,6 +692,7 @@ check(PERMISSIONS.IOS.LOCATION_ALWAYS)
     switch (result) {
       case RESULTS.UNAVAILABLE:
         console.log('This feature is not available (on this device / in this context)');
+        console.log('This can happen e.g. if "Privacy" => "Location Services" is turned off')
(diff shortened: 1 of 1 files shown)


zoontek:
@miallo The current explanation is not incorrect. As it's in the `check` method documentation, we need a generic message, the location permission is used as an example, but the method work for every permissions.

This case fits in:

> **This feature is not available** (on this device / **in this context**)

miallo:
Thank you very much for this great library!
Yes, I totally agree that it was correct before and I probably was just a bit stupid when reading the docs...
Have a nice day!

PR #687, opened 2022-05-19
Add support for iOS Local Network permission

### LINKED ISSUE #509: iOS14 - Local Network permission support
# Feature request

Support for Local Network permission needed (iOS14)

## Why it is needed
With iOS 14 user may disallow Local Network permission for mobile app, which will deny access for app to local devices. 
https://developer.apple.com/videos/play/wwdc2020/10110/

### Should be used like something like this:

```
    Permissions.getPermissionStatus('localNetwork').then((r) => {
        if (r === 'authorized') {
           // Do something
        } else if (r === 'undetermined') {
          Permissions.requestPermission('localNetwork').then((response) => {
              if (response === 'authorized') {
                           // Do something
              }
              if (response === 'denied') {
               // Notify user about Local Network permission denied, advice user to turn on permission
               }
            });
        } else {
               // Notify user about Local Network permission denied, advice user to turn on permission
        }
      });
```

geraintwhite:
# Summary

Fix #509

## Test Plan

Tested in my app by calling the following and verifying that the permission prompt is displayed:

```
Permissions.check(PERMISSIONS.IOS.LOCAL_NETWORK_PRIVACY)
Permissions.request(PERMISSIONS.IOS.LOCAL_NETWORK_PRIVACY)
```

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

- [x] I have tested this on a device and a simulator (local network privacy not available on simulator)


README.md
```diff
@@ -47,12 +47,14 @@ target 'YourAwesomeProject' do
 
   permissions_path = '../node_modules/react-native-permissions/ios'
 
+  pod 'RNPermissions', :path => '../node_modules/react-native-permissions', :modular_headers => true
```
mikehardy: this one is, I think, auto-linked and not necessary? Does something about this PR make it necessary now?

example/ios/RNPermissionsExample/Info.plist
```diff
@@ -35,6 +35,10 @@
 			</dict>
 		</dict>
 	</dict>
+	<key>NSBonjourServices</key>
+	<array>
+		<string>_lnp._tcp.</string>
+	</array>
```
mikehardy, line 41: this  may need a separate documentation chunk to avoid a million "how do I use this?" questions :-)

ios/LocalNetworkPrivacy/Permission-LocalNetworkPrivacy.podspec
```diff
@@ -0,0 +1,20 @@
...
+  s.dependency                  "RNPermissions"
+
+  s.version                   = package["version"]
+  s.license                   = package["license"]
+  s.summary                   = package["description"]
+  s.authors                   = package["author"]
+  s.homepage                  = package["homepage"]
+
+  s.ios.deployment_target     = "10.0"
+  s.tvos.deployment_target    = "11.0"
```
mikehardy, line 15: I have not personally checked, but have you checked all APIs are actually available on ios and tvos? out of curiosity, are there any restrictions for ipados / macCatalyst or macos? (just thinking cross-platform, though it's not a strict requirement or anything)

mikehardy (commented):
I think this is neat so I'm leaving comments with the intention of being helpful but I haven't personally investigated them (apologies) or pulled this and tried it (apologies again), and @zoontek is the repo boss, so it's all up to him anyway. Just wanted to see if I could help move it along. Cheers

ios/LocalNetworkPrivacy/Permission-LocalNetworkPrivacy.podspec
geraintwhite: Haven't checked yet, just based the podspec on one of the others.

example/ios/RNPermissionsExample/Info.plist
geraintwhite: Good shout, will add

README.md
geraintwhite: This was needed to get around an error on pod install caused by mixing Swift and Objective C (the modular_headers is the important bit) - there may be another way around it

zoontek:
Thanks a lot for this one! However, is it possible for you to migrate the module to ObjC? There's already too much languages on this repository (ObjC, Java, TypeScript, C#…).

ios/LocalNetworkPrivacy/Permission-LocalNetworkPrivacy.podspec
geraintwhite: According to the docs of [NetService](https://developer.apple.com/documentation/foundation/nsnetservice) it is available on all platforms - but it has been deprecated with no obvious replacement.

example/ios/RNPermissionsExample/Info.plist
geraintwhite: Added in b5701475468a1f89e62b62a2f4ab5991a91691d6

README.md
geraintwhite: Removed now I've rewritten the handler in objective c d2d598fc5aa0bcfaf0aa2c186db65207f9061a01

geraintwhite:
> Thanks a lot for this one! However, is it possible for you to migrate the module to ObjC? There's already too much languages on this repository (ObjC, Java, TypeScript, C#…).

Converted to objective c and tested it still works d2d598fc5aa0bcfaf0aa2c186db65207f9061a01

ios/LocalNetworkPrivacy/Permission-LocalNetworkPrivacy.podspec
mikehardy: Oh that's unfortunate. And you're right, no real guidance. You probably found this as well but the only thing I found was the lower level stuff here https://developer.apple.com/library/archive/samplecode/DNSSDObjects/Introduction/Intro.html#//apple_ref/doc/uid/DTS40011371-Intro-DontLinkElementID_2 - everything links to it

geraintwhite:
It's occurred to me that checking the permission will only return an accurate result **after** requesting it, as the only way to check it is by requesting it first and storing the result (https://stackoverflow.com/questions/63940427/ios-14-how-to-trigger-local-network-dialog-and-check-user-answer/64242745#64242745).

I can think of three possible solutions:
- leave it as is so the check result will be inaccurate until it is requested
- disable `Permissions.check` for this permission
- persist the result somewhere other than a class property.

zoontek:
@grit96 The library has a builtin escape hatch for these cases: https://github.com/zoontek/react-native-permissions/blob/37d732e8333f1a653300292e82d08decaf9e3995/ios/RNPermissions.m#L363

Get `isFlaggedAsRequested` value on check. If it's false, return "NotDetermined". If it's true, you can perform a request to get the status (don't forget to call `flagAsRequested` on request).

You can have an usage example in the FaceID handler: https://github.com/zoontek/react-native-permissions/blob/2d645ff4f3dfc6db9686bc2fd2877bd049b8043e/ios/FaceID/RNPermissionHandlerFaceID.m

geraintwhite:
@zoontek something like this?

```objc
#import "RNPermissionHandlerLocalNetworkPrivacy.h"
#import "LocalNetworkPrivacy.h"

@implementation RNPermissionHandlerLocalNetworkPrivacy

+ (NSArray<NSString *> *_Nonnull)usageDescriptionKeys {
  return @[ @"NSLocalNetworkUsageDescription" ];
}

+ (NSString *_Nonnull)handlerUniqueId {
  return @"ios.permission.LOCAL_NETWORK_PRIVACY";
}

- (void)checkWithResolver:(void (^_Nonnull)(RNPermissionStatus))resolve
                 rejecter:(void(__unused ^ _Nonnull)(NSError *_Nonnull))reject {
  if (![RNPermissions isFlaggedAsRequested:[[self class] handlerUniqueId]]) {
    return resolve(RNPermissionStatusNotDetermined);
  }

  LocalNetworkPrivacy *local = [LocalNetworkPrivacy new];
  [local checkAccessState:^(BOOL granted) {
    resolve(granted ? RNPermissionStatusAuthorized : RNPermissionStatusDenied);
  }];
}

- (void)requestWithResolver:(void (^_Nonnull)(RNPermissionStatus))resolve
                   rejecter:(void (^_Nonnull)(NSError *_Nonnull))reject {
  [RNPermissions flagAsRequested:[[self class] handlerUniqueId]];
  [self checkWithResolver:resolve rejecter:reject];
}

@end

```

zoontek:
@grit96 Exactly. With your current request code:

```objc
- (void)requestWithResolver:(void (^ _Nonnull)(RNPermissionStatus))resolve
                   rejecter:(void (^ _Nonnull)(NSError * _Nonnull))reject {
  LocalNetworkPrivacy *local = [LocalNetworkPrivacy new];
  [local checkAccessState:^(BOOL granted) {
      [RNPermissions flagAsRequested:[[self class] handlerUniqueId]];
      [self checkWithResolver:resolve rejecter:reject];
  }];
}
```

geraintwhite:
I've moved the network request from requestWithResolver into checkWithResolver so it doesn't get called twice.

zoontek:
I recommend the opposite : perform a request in check (for consistency)

nilrezaee88:
Hi 
i used 
Permissions.check(PERMISSIONS.IOS.LOCAL_NETWORK_PRIVACY)
this error generated  -> Cannot find name 'Permissions'. Did you mean 'PERMISSIONS'?
i need Network privacy permission check and displaying message for accessing application to Local Network
please help me
tnx a lot
@zoontek 
@grit96

bintoll:
Hello! Thanks for this PR, I've tested this and find out that straight after permission request it immediately returns BLOCKED as a response while user request alert is still on screen and user did not allowed or denied it, so it just do not wait untill user will process with any action

geraintwhite:
@bintoll hmm you are correct - it's happening because there is no way to detect when the user presses decline, so we have a timer which automatically returns denied after 2s. There does seem to be an alternative using NWBrowser but I can't find any objc equivalent API (https://stackoverflow.com/a/67758105/2468126).

bintoll:
@grit96 Got it, that generally is fine, it will not lead to the best UX for user (in case if we would like to alert user after declining permission and allow him to navigate to device settings), but requesting permission itself works nice. 
I would recommend to mention this behavior somewhere in readme so the flow would be clear and developers would not create issues regarding this functionality. 
Thank you again for implementing this!

mikehardy:
after a decline you can re-check (but not re-request!) in order to differentiate perhaps? I have not looked that closely though, if that idea doesn't fit at all, apologies

bintoll:
@mikehardy It seems that it could work, it will be needed to setup an interval to check status consistently until the status will be changed (when user finally will press allow or decline). Will check it al little bit later. 
But anyway, as long as this flow breaks the common behavior of requesting permissions in this lib, it's better to mention this.

nilaydigiwhiz:
Why this PR is not merged yet?

zoontek:
@nilaydigiwhiz https://jacobtomlinson.dev/posts/2022/dont-be-that-open-source-user-dont-be-me/

I do almost free, volunteer work on this library (and others) and have a full time job, that's why.

Also, every time someone say things like "Why is this not done yet?", "My company need this!" (for free, obviously) or else, my motivation doesn't go up, isn't it weird? 😄

Anyway, I'm currently on holidays ☀️ and will probably take some time from it to review it / test it on several devices and simulators / release it (and later, maintain it and reply to issues - just in case you though that being an open source maintainer is just pressing "merge" buttons)

vitalyiegorov:
@zoontek Thanks for your effort and for this great library! Did you have time to review/test this PR? Maybe I can help in someway with this?

zoontek:
@vitalyiegorov Not yet.

FC04FED5-4F72-4651-AB62-6D5786A53B14

Sure, you can try and confirm if it works well.

zoontek:
I took a check and finally try the @grit96 solution. IMHO this is not enough (and I'm afraid it can't be better)
The permission must be requested using the library or it will fails on check (ex: here react native in dev mode will prompt for permission at app start)

Also, this indeed use deprecated APIs so it's not future proof. I'd rather not merge this as it will create confusions and issues.

https://user-images.githubusercontent.com/1902323/189542881-c009ee70-e0c6-49c3-aa16-241dab03e455.MP4

ios/LocalNetworkPrivacy/LocalNetworkPrivacy.m
```diff
@@ -0,0 +1,47 @@
...
+        self.service = [[NSNetService alloc] initWithDomain:@"local." type:@"_lnp._tcp." name:@"LocalNetworkPrivacy" port:1100];
+    }
+    return self;
+}
+
+- (void)dealloc {
+    [self.service stop];
+}
+
+- (void)checkAccessState:(void (^)(BOOL))completion {
```
dylancom, line 26: Is this as reliable as:
(It seems you altered the code from: https://stackoverflow.com/a/65800221)
```
    func checkAccessState(completion: @escaping (Bool) -> Void) {
        self.completion = completion
        
        timer = .scheduledTimer(withTimeInterval: 2, repeats: true, block: { timer in
            guard UIApplication.shared.applicationState == .active else {
                return
            }
            
            if self.publishing {
                self.timer?.invalidate()
                self.completion?(false)
            }
            else {
                self.publishing = true
                self.service.delegate = self
                self.service.publish()
                
            }
        })
    }
```

Audrey-Ann:
Any updates here?

idoodler:
Any Updates here?

ngandhy:
To note @geraintwhite - the apple docs were rewritten Oct 2024 with how to check status / denied / etc.. so hoping that with some tweaks it will handle @zoontek 's concerns around merging this.

this is especially relevant with Android 16 adding a similar permission that folks will likely want to be handled via this library as well.

mikehardy:
> The simulator doesn’t support local network privacy. Test your local network privacy behavior on a real device.

😆  ohhhh Apple.

Issue #688, opened 2022-05-27
Add support for Android 13 Notification permission


fabiendem:
### Why it is needed?

Android 13 will ship with a new permission: Notification permission.
https://developer.android.com/about/versions/13/changes/notification-permission

### Possible implementation

I suppose it could fit in the existing API `check/requestNotifications`

### Code sample

```js
Sorry I don't have a snippet for now
```

mikehardy:
We're working on this already for Notifee and have a PR queued but are waiting to make sure Tiramisu / 13 stabilizes before landing. It's nothing fancy though. Just cross-pollinating in here in case it is useful at all

```java
public void requestPermission(Activity activity, MethodCallResult<Bundle> result) {
    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
      requestPermissionCallResult = result;
      ActivityCompat.requestPermissions(activity, new String[]{Manifest.permission.POST_NOTIFICATIONS}, REQUEST_CODE_NOTIFICATION_PERMISSION);
    } else {
      getNotificationSettings(result);
    }
  }

  public boolean onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
    if (requestCode == REQUEST_CODE_NOTIFICATION_PERMISSION) {
      if (requestPermissionCallResult != null) {
        getNotificationSettings(requestPermissionCallResult);
        return true;
      }
    }
    return false;
  }
```

https://github.com/invertase/notifee/pull/351/files#diff-f5d7290ccfe968a3611168b4c271053d229241b15675d95bafdf35c32b3f5f23R424-R441

zoontek:
This feature has landed in [v3.5.0](https://github.com/zoontek/react-native-permissions/releases/tag/3.5.0)

fabiendem:
Merci!

Issue #684, opened 2022-04-09
[call for sponsors] - React Native new architecture support


zoontek:
Hi everyone 👋

@zoontek here, creator and maintainer of `react-native-{bootsplash / localize / permissions / bars / dev-menu}`.
I'm opening this issue to prevent multiple ones from being created about the new React Native architecture compatibility ([introduced in version `0.68`](https://reactnative.dev/blog/2022/03/30/version-068)).

I'm excited about these changes and the future of React Native, but it's quite hard for me to free up time to work on it.
To explain myself a little bit: I did a lot of open source these last years, mainly for free - and I enjoyed it, don't even doubt about that 😄. It's not only about writing code, but also about helping people, so they can craft great products ✨.

But time flies. On a personal note, I'm currently trying to buy my own flat and started a freelance activity in addition to my regular job to achieve this goal.
Open source is time-consuming (handling issues, reproducing, reviewing PRs, etc) and doesn't pay bills. Even worse: it currently costs me money (I bought several used phones for testing, a personal MacBook recently). I could use this time to work for paid freelancing instead.

On the other side, every day I see companies using these libraries in quite popular apps. That's why I'm calling them for sponsorship. It doesn't have to be a recurrent, monthly sponsorship, but if you want to push for the new arch landing on this repository, understand that it will be easier for me to find some time if I'm being paid for the task 🙂 (actually, it could be fast!)

👉 For that, I created one-time sponsorship tiers ([available here](https://github.com/sponsors/zoontek?frequency=one-time)). You don't have to respect the amounts (custom is possible - maybe you are an individual and can only give a little or maybe you want to give even more than the last tier).
If you are not used to giving money for open source software, you can think of it as a one-time payment license (even if it's clearly not - it will stay free and MIT).

P.S. I'm aware I could also wait for a PR written for free by a community member to land, but that's **not** what I want.
Even if it happens, I will still have to learn about the new arch, review it, and will maintain it / answer your questions after it's merged. So it will probably be easier if I work on it initially.

P.S. (2) That doesn't mean that it will never land if I don't get sponsorships, just that it will be less of a priority for me.

BraveEvidence:
This can help https://www.youtube.com/watch?v=Bn-HIVb2ZIw

fobos531:
@zoontek I'm assuming this can be closed now that #748  is merged?

PR #696, opened 2022-06-25
3.4.0 beta (supports Ask next time, Only this time, etc)

### LINKED ISSUE #565: Returns Block in android even if the permission is granted in the settings
Returns Block in android even if the permission is granted in the settings !!!!!!!!

## Environment info
Library version: 
    "react": "16.13.1",
    "react-native": "0.63.4",
    "react-native-permissions": "^3.0.0",

## Steps to reproduce
1. first，blocked the permission ；
2. then, and open the permission in system settings;
3. last,check or request the permission , but the result is always BLOCKED.

Describe what you expected to happen:

 1. after open the permission in system settings, then i check or request the permission, the result should not BLOCKED!!!!!

## Reproducible sample code

    check(permission)
      .then(result=>{
        console.log("check The permission is "+result);
        switch (result) {
          case RESULTS.UNAVAILABLE:
          case RESULTS.LIMITED:
          case RESULTS.GRANTED:
          case RESULTS.BLOCKED:
            callback(result.toString());
            break;
          case RESULTS.DENIED:
            request(permission).then(ret => {
              console.log("request The permission is "+result);
              if (ret===RESULTS.GRANTED){
                callback(ret.toString());
              }else {
                console.log("The permission is "+ret);
              }
            }).catch(err => {
              console.log("err " + err.toString());
            });
            break;
        }
      }).catch(err=>{
      console.log("err " + err.toString());
    });

zoontek:
Hi everyone 👋

In order to fix [this irritating issue](https://github.com/zoontek/react-native-permissions/issues/565) and others like "Only this time" permission grant on Android >= 12, I decided to remove `SharedPreferences` usage in the Android part of the module. Which mean that `check()` could now only respond with `RESULTS.UNAVAILABLE`, ` RESULTS.GRANTED` or `RESULTS.DENIED` and you cannot be sure that the permission isn't `RESULTS.BLOCKED` until you request it. And even after that, the `check` method will continue to return `RESULTS.DENIED`.

This is far from perfect, but is related to how the Android permission works. See [this comment](https://github.com/zoontek/react-native-permissions/issues/565#issuecomment-870628975) to understand it better.

### Try it!

You can try it using `yarn add react-native-permissions@next` and give your feedbacks ✨
For that, I also updated the example app (on [this branch](https://github.com/zoontek/react-native-permissions/tree/3.4.0)):


zoontek:
Note that I will also update the Android flow in the README to help everyone keeping an OK UX.

zoontek:
This version will reintroduce this error: https://github.com/facebook/react-native/issues/30158

Quite a regression, but there is currently no way to correctly detect a `BLOCKED` status on Android (checking `shouldShowRequestPermissionRationale` is a hack, and does not work when you dismiss the popup on Android >= 11).

The Android permission system is broken as its core and nobody at Google seems to care 🤷🏻‍♂️

zoontek:
I updated the Android flow chart:

Screenshot 2022-06-26 at 11 21 53

What I recommend to keep a "good" UX on Android:

1. When arriving on a screen where a permission is required, check it.
2. It could be `GRANTED` (use the feature immediately) or `DENIED`
3. If it's denied, display a screen with some explanations on why it's needed + a button to request
4. Request the permission. If it's `BLOCKED`, replace the content + the button with "Go to settings" instructions

It's not perfect for two reasons:

- It could be a lot better to know if the permission is `BLOCKED` and display "Go to settings" directly at start, but we cannot do that without relying on storage and it messes with "Only this time" grants, or "Ask next time" feature
- Dismissing the request popup will result on `BLOCKED` and it will switch to the "Go to settings" screen, even if the permission is still requestable.

But I'm afraid we can't achieve better.

artyorsh:
Hi @zoontek,
Any chance we can also include #688 in the upcoming version?

zoontek:
@artyorsh Hi! No plan for now since I still struggle a bit to get [company sponsorships](https://github.com/zoontek/react-native-permissions/issues/684). If your company need it quickly, you can contact me and I will put high priority on it 🙂

artyorsh:
@zoontek thanks for the quick response! I respect your motivation with the sponsorships - the permissions library is definitely worth it 🙂
 
For the Android 13 topic - most of the apps will have to upgrade to a new notification permission requirement (which doesn't necessarily involve new arch migration). Given the fact that it's available in [Platform Stability](https://android-developers.googleblog.com/2022/06/android-13-beta-3-platform-stability.html), can we have a look on [what other libraries do](https://github.com/invertase/notifee/pull/351) and raise the topic also in this library? I would be happy to contribute with moving the code from there to this package and checking it on emulators but I don't have a physical Android 13 device

zoontek:
@artyorsh I will update the call for sponsors post. Indeed it's not related to the new architecture, but we need a way to make libraries development sustainable.
Currently there's a lot of feature requests on this repository, I need to found a proper way to handle that (aka, not for free - companies need to understand that it will cost them a lot more to save a few bucks 😅)

vincent-paing:
I saw that my PR for notifee is mentioned. I'll be happy to assist with adding newly introduced permission in Android 13

Irenedan:
> This version will reintroduce this error: [facebook/react-native#30158](https://github.com/facebook/react-native/issues/30158)
> 
> Quite a regression, but there is currently no way to correctly detect a `BLOCKED` status on Android (checking `shouldShowRequestPermissionRationale` is a hack, and does not work when you dismiss the popup on Android >= 11).
> 
> The Android permission system is broken as its core and nobody at Google seems to care 🤷🏻‍♂️

@zoontek Hi, So how could you avoid this kind of problem in user experience?
At present, I will pop up a popup to remind the user when I receive the never_ask_again status, but this bug causes the popup to pop up when dismisses.

zoontek:
@Irenedan You can't. Previous versions of `react-native-permissions` relied on `SharedPreferences` storage, but it introduced other issues on Android 11+

To understand the issue with the Android permission system here:

Before Android 11, a permission request couldn't be dismissed, we have to grant or denied it. On the second try, the user had a checkbox "Never ask again". As after the first try, android `shouldShowRequestPermissionRationale` method returns `true` is the user denied the request, we could be sure that it was `DENIED`, otherwise, it was `NEVER_ASK_AGAIN`. Easy.

But since Android 11, as the permission request can just be dismissed, if `shouldShowRequestPermissionRationale` returns `false`, is it because:

1) The user tick "Never ask again" / denied twice (the new system)?
2) The user dismissed the popup?

As the app briefly lose focus even if no popup appear, the only crappy hack that come to my mind would be to measure how many time it took to the user to deny. Far from perfect, but until Google fixes this we have no other choices than hacks like that. And this king of stuff couldn't be included in this library since it would not work with `requestMultiple`

More infos: https://developer.android.com/training/permissions/requesting

Irenedan:
Hello, I tried to use 'yarn add react-native-permissions@next', but the app crashes when the user selects the option 'Only this time'

Phone: One plus with Android 12

zoontek:
@Irenedan Use `3.4.0`, this PR has been merged

Edit: Just tried it, no issue from the library (Google Pixel 4a, android 12):

mikehardy:
>  the app crashes

This phrase should always be followed with "...and here's the stack trace from `adb logcat`" :-) https://developer.android.com/studio/command-line/logcat

PR #726, opened 2022-10-21
Improve `request` method documentation

fabiendem:
# Summary

The `request` method takes an optional `rationale` parameter.  

This parameter is read on **Android only**:

https://github.com/zoontek/react-native-permissions/blob/6cd97d514b01f5655ac92c1f13352af0e4a05724/src/methods.android.ts#L36

It is not read on iOS: 
https://github.com/zoontek/react-native-permissions/blob/6cd97d514b01f5655ac92c1f13352af0e4a05724/src/methods.ios.ts#L41

Neither Windows:

https://github.com/zoontek/react-native-permissions/blob/6cd97d514b01f5655ac92c1f13352af0e4a05724/src/methods.windows.ts#L26

This PR tweaks the documentation to make it clear that the `Rationale` is only useful on Android.

Happy to tweak the wording if needed.

## Test Plan

No test plan, see updated `README`.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅    |

## Checklist

- [] I have tested this on a device and a simulator
- [] I added a sample use of the API in the example project (`example/App.tsx`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -738,7 +738,7 @@ check(PERMISSIONS.IOS.LOCATION_ALWAYS)
 
 #### request
 
-Request one permission.  
+Request one permission.
 
 The Rationale is available and read on Android only.  
 On iOS, update your `Info.plist` with wanted permissions usage descriptions.
--- a/README.md
+++ b/README.md
@@ -740,8 +740,7 @@ check(PERMISSIONS.IOS.LOCATION_ALWAYS)
 
 Request one permission.
 
-The Rationale is available and read on Android only.  
-On iOS, update your `Info.plist` with wanted permissions usage descriptions.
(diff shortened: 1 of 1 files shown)


zoontek:
@fabiendem Hi 👋 Thanks for the contribution!

I think having:

> Note that the `rationale` parameter is only available and used on Android.

Will be more clear. No need to mention iOS here since rationales != prompt texts. What do you think?
PS: Don't forget to run prettier on the markdown or to remove the trailing spaces 🙂

fabiendem:
Hi @zoontek !
No probs, I have tweaked the wording there.
Format shouuuuld be ok now.

zoontek:
Yep, it's perfect, thanks!

PR #716, opened 2022-09-16
Handle upgrading from location when in use to location always on iOS

tallpants:
# Summary

* Closes https://github.com/zoontek/react-native-permissions/issues/490
* Fixes an issue on iOS where you could not ask for `LOCATION_ALWAYS` permission if you already had `LOCATION_WHEN_IN_USE` permission.
* This only impacts the `LocationAlways` pod.

## Implementation Changes

- For `check`:
  - If the current authorization status is `AuthorizedWhenInUse`, and we _have not_ requested `LOCATION_ALWAYS` in the past, we return `NotDetermined`.
  - If the current authorization status is `AuthorizedWhenInUse` and we _have_ requested `LOCATION_ALWAYS` in the past, we return `Denied`, since we're not allowed to prompt for this permission twice.
- For `request`:
  - If the current authorization status is `AuthorizedWhenInUse` and we _have not_ requested `LOCATION_ALWAYS` in the past, we call `requestAlwaysAuthorization` and flag the permission as requested.
  - If the current authorization status is `AuthorizedWhenInUse` and we _have_ requested `LOCATION_ALWAYS` in the past, then we return the current authorization result, since we're not allowed to prompt for this permission twice.

## Test Plan

- Request and grant `LOCATION_WHEN_IN_USE` permission.
- Then request `LOCATION_ALWAYS` permission.

| Before | After |
| --- | --- |
| Before | After |

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    N/A   |

## Checklist

- [X] I have tested this on a device and a simulator
- [X] I added the documentation in `README.md`
- [X] I added a sample use of the API in the example project (`example/App.tsx`)

PR #734, opened 2022-12-02
fix: allow LocationAlways permission request when current status is kCLAuthorizationStatusAuthorizedWhenInUse

ag-drivequant:
Same issue explained at https://github.com/zoontek/react-native-permissions/pull/716

The fix let the module accept requesting for LocationAlways permission, once we have previously requested the LocationWhenInUse permission.

This scenario is not working on the current master branch.


zoontek:
Duplicate of https://github.com/zoontek/react-native-permissions/pull/716

The issue is more complex than that.

Issue #632, opened 2021-07-16
Add android permission REQUEST_IGNORE_BATTERY_OPTIMIZATIONS


ghost:
# Feature request

It would be great, if you could add the check and request permission for 'android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS'.

## Why it is needed

Our app needs the permission to run an infinite background service on android. We implemented a working react native module for testing the code mentioned below. But it would be cool, if it would be part of your library.

## Possible implementation

https://stackoverflow.com/questions/55794316/action-ignore-battery-optimization-settings-need-intent-to-all-apps-list

zoontek:
Duplicate of https://github.com/zoontek/react-native-permissions/issues/395
@felix4321 Nobody did a PR, could you?

ghost:
@zoontek Thx for info - I will try the PR!

mikehardy:
@felix4321 careful with this one! has some edge cases apparently where Activity may not exist, and it may cause Play Store rejection https://stackoverflow.com/a/50840603/9910298

It's not a dangerous permission to start, so I think this is more about just opening the settings after giving the user rationale? 

In that case, I think it exists: https://github.com/rasheedk/react-native-disable-battery-optimizations-android

And requesting permission here does not actually make sense, because it is not some restricted permission that the user has to grant, rather it is a settings toggle you need to explain etc

zoontek:
After checking, this is indeed too risky to be put in this library. As it's something exceptional to request this kind of permissions, I think having a second library that only do that makes sense.

PR #737, opened 2022-12-07
Battery Permissions Functionalities

Gabriel-KarkG:
# Summary

When the permission `REQUEST_IGNORE_BATTERY_OPTIMIZATIONS` is put in the Android Manifest, the `requestPermission` method would always return `true` for this permission. The `checkPermission` method now instead checks if the device is ignoring battery optimization, and returns `true` if it is, or `false` if it isn't.

Additionally, the method `triggerBatteryOptimizationNativeDialog` was also added to be able to trigger the permission dialog that asks the user to disable the battery optimization setting for the app. 

## Test Plan

https://user-images.githubusercontent.com/97614211/206280711-8fb137b4-7795-42c4-a502-e5439d518610.mp4

https://user-images.githubusercontent.com/97614211/206280721-1f4de73c-5353-4f65-aaf6-c370f29ea2a8.mp4

NA
NA
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/zoontek/rnpermissions/RNPermissionsModule.java
+++ b/android/src/main/java/com/zoontek/rnpermissions/RNPermissionsModule.java
@@ -203,6 +203,17 @@ public class RNPermissionsModule extends ReactContextBaseJavaModule implements P
       return;
     }
 
+    if (permission.equals("android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS")){
+      String packageName = context.getPackageName();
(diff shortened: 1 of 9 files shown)


zoontek:
Hi! Thanks for the PR.

This feature [already has been discussed](https://github.com/zoontek/react-native-permissions/issues/632#issuecomment-881597706)  on this repository and we think it's not a good idea to provide it.

Gabriel-KarkG:
All good, appreciate the feedback!

PR #750, opened 2023-02-15
Add react-native setup-ios-permissions command

zoontek:
Hi folks 👋

Let's face it: the [current iOS permission handlers setup](https://github.com/zoontek/react-native-permissions/tree/3.6.1#ios) is not that great. Having to deal with the `Podfile` file is a chore and not well understood by "non-native" developers.
This pull request adds a new system based on a new `react-native` CLI command: `setup-ios-permissions`. To use it, it's fairly easy!

1️⃣ First declare the used iOS permissions in your `reactNativePermissionsIOS` config (the command uses [`cosmiconfig`](https://github.com/davidtheclark/cosmiconfig), so we can put that in our project `package.json` if we want):

```json
{
  "reactNativePermissionsIOS": [
    "Camera",
    "LocationWhenInUse",
    "Notifications",
    "SpeechRecognition"
  ]
}
```

2️⃣ Then run the CLI command:

```bash
$ npx react-native setup-ios-permissions
```

3️⃣ Don't forget to reinstall Pods:

```bash
$ npx pod-install
```

4️⃣ And…it's done ✨

P.-S. Note that you can use a `postinstall` script to simplify this:

```json
{
  "scripts": {
    "postinstall": "react-native setup-ios-permissions && pod-install"
  },
  "devDependencies": {
    "pod-install": "0.1.38"
  },
  "reactNativePermissionsIOS": [
    "Camera",
    "LocationWhenInUse",
    "Notifications",
    "SpeechRecognition"
  ]
}
```

P.-S. (2) Did I mention that this approach fixes the issue with `use_frameworks!` that need [this ugly workaround](https://github.com/zoontek/react-native-permissions/tree/3.6.1#workaround-for-use_frameworks-issues)? 👀 (As permissions handlers will now be parts of the `RNPermissions` pod directly):

pods


mikehardy:
Fascinating!! Very nice (as probably the creator of all the motivation for use_frameworks adoption in react-native 😅 )

react-native.config.js
```diff
@@ -0,0 +1,61 @@
...
+        if (!Array.isArray(config)) {
+          console.log(pc.red('Invalid config'));
+          process.exit(1);
+        }
+
+        if (config.length === 0) {
+          console.log(pc.yellow('Empty config'));
+        }
+
+        const iosDirPath = path.join(__dirname, 'ios');
```
thymikee, line 38: This path can be different in monorepos and exotic setup. RN cli plug-ins have access to config object part of fn arguments) where you can get "config.project.ios.sourceDir"

zoontek: @thymikee I know, but this is not the project `ios` directory here, but `node_modules/react-native-permissions/ios` 🙂

thymikee: Ugh, I didn't realize when checking on a mobile 😅

thymikee:
I wonder about the name in package.json. `"iosPermissions"` is pretty generic and doesn't immediately point to this library. Since this is probably gonna be copy-pasted initially, how about something more explicit like `"reactNativePermissionsIOS"`? On the other hand, what's the chance other nodejs tool would use this namespace? Probably really low. Thoughts inspired by https://github.com/react-native-community/discussions-and-proposals/pull/588

zoontek:
@thymikee Good though! I updated the config key.

helenaford:
@zoontek I appreciate the library and all the work you put in 🙏  just struggling a little to see how this makes it easier or more performant. Sometimes you just want to run `yarn install`  to update the js node modules without running pod install everytime. 

For some projects im working on, they use expo, so this is a very common scenario where pods aren't needed when running `yarn install` .

Any way to have a solution where the pods are nicely structured but without the need for an setup script?  (cc @mikehardy ;) )

zoontek:
@helenaford You actually don't have to run `pod install` each time you run a `yarn install`. I added that in README so people don't forget it.

You can remove the `pod install` from `postinstall` if you want, just don't forget to run it manually every time you change the `reactNativePermissionsIOS` config / every time you delete the `node_modules` directory  or run `yarn install --force`

helenaford:
@zoontek yeah sorry, that's what i meant, every time the node modules are removed, you have to run the setup script right? 

So for CI, we need the postinstall step?

helenaford:
also, for local development, i'd say it's not great to have that manual step. I can see how that can easily get missed out each time the node modules are cleared

zoontek:
@helenaford Yes. The CLI modify the podspec source paths, then you run the pod install.

Not ideal, but unfortunately the best compromise we found yet 😕

helenaford:
@zoontek  like i said above, i really appreciate how hard this is to do and what the intention is to try and make it as easy as possible, but in my opinion, the other way was better because you just configured it once per permission module.  Rather than everytime the node modules are cleared.

is there a way to have two methods to setup the library?

zoontek:
@helenaford The other way is still possible (the podspecs per handler are still there), but will probably be removed in next major version as it creates way too much issues with Xcode caching.

Maybe we could add a similar command for the Pod file? Something like `use_react_native_permissions!`, able to read the `reactNativePermissionsIOS` too? I will investigate.

helenaford:
@zoontek thank you that would be perfect if we could put something in the Podfile. thanks for looking into this! 😎

zoontek:
@helenaford Could you try it? I released a new version, `3.9.0-beta.0` with it.

```sh
yarn add react-native-permissions@next
```

Then update your `Podfile` like this:

```diff
require_relative '../node_modules/react-native/scripts/react_native_pods'
require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'
+ require_relative '../node_modules/react-native-permissions/scripts/permissions_setup'

platform :ios, min_ios_version_supported
prepare_react_native_project!
+ prepare_react_native_permissions!
```

And run `pod install` (`pod install` must be re-executed each time you update `reactNativePermissionsIOS` config, but that's all)

**PS:** Don't forget to remove the `pod-install` call on `postinstall` if you want, as it's not needed anymore.

If you are curious, this is the script: https://github.com/zoontek/react-native-permissions/blob/add-ruby-setup-script/scripts/permissions_setup.rb

Basically a conversion of the react native CLI plugin to ruby. Thanks again for the insight!

---

I'm not sure about the wording (the script file + function name) tho, it is subject to change before a release. Does it sounds good for you? @mikehardy @thymikee, do you have an opinion / some ideas for that?

Note that putting everything in the `Podfile` (no `reactNativePermissionsIOS` config anymore), could also be a solution. It might even be simpler:

```diff
require_relative '../node_modules/react-native/scripts/react_native_pods'
require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'
+ require_relative '../node_modules/react-native-permissions/scripts/permissions_setup'

platform :ios, min_ios_version_supported
prepare_react_native_project!

+ setup_permissions([
+  'Camera',
+  'LocationWhenInUse',
+  'Microphone'
+ ])
```

zoontek:
I added a commit with it (could be rolled back): https://github.com/zoontek/react-native-permissions/commit/51d97fa5cdc17a3481e5bc49b19ba7b0e7fc4afd. This looks neat, if you ask me 😄. And as the user will have to update his `Podfile` to add / delete a permission, having to run a `pod install` will make total sense for him.

EDIT: I published it as `3.9.0-beta.1`, as I find this solution much more elegant.

PR #755, opened 2023-03-07
update NotificationOption type

ruslan-hontar:
# Summary

PR includes an update for `NotificationOption` type

`checkNotifications` returns `settings` based on the `NotificationSettings` type, but we can't request the full list option in the `requestNotifications` because of `NotificationOption` type.
i tried to pass missed `lockScreen` and  `'notificationCenter'` to the `requestNotifications` method and  get needed permissions. i think, they were missed and not included in some of pr

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -793,7 +793,9 @@ type NotificationOption =
   | 'criticalAlert'
   | 'carPlay'
   | 'provisional'
-  | 'providesAppSettings';
+  | 'providesAppSettings'
+  | 'lockScreen'
+  | 'notificationCenter';
 
 type NotificationSettings = {
   // properties only available on iOS
--- a/src/types.ts
+++ b/src/types.ts
@@ -27,7 +27,9 @@ export type NotificationOption =
   | 'carPlay'
   | 'criticalAlert'
   | 'provisional'
-  | 'providesAppSettings';
+  | 'providesAppSettings'
+  | 'lockScreen'
+  | 'notificationCenter';
 
 export type NotificationSettings = {
   alert?: boolean;


zoontek:
Hi 👋

Thanks for the pull request, but actually, everything is OK.

- `lockScreen` and `notificationCenter` are not valid [`UNAuthorizationOptions`](https://developer.apple.com/documentation/usernotifications/unauthorizationoptions?language=objc) (which mean [they couldn't be requested](https://github.com/zoontek/react-native-permissions/blob/master/ios/Notifications/RNPermissionHandlerNotifications.m#L87))
- `requestNotifications` [argument](https://github.com/zoontek/react-native-permissions/blob/master/src/methods.ios.ts#L60) is indeed [`NotificationOption[]`](https://github.com/zoontek/react-native-permissions/blob/master/src/types.ts#L23)
- `checkNotifications` and `requestNotifications` [return type](https://github.com/zoontek/react-native-permissions/blob/master/src/methods.ios.ts#L61) is [`NotificationsResponse`](https://github.com/zoontek/react-native-permissions/blob/master/src/types.ts#L44)
- [`NotificationSettings`](https://github.com/zoontek/react-native-permissions/blob/master/src/types.ts#L32) (in `NotificationsResponse`) correctly include `lockScreen` and `notificationCenter` has those values could be **checked**

ruslan-hontar:
thanks for the explanations

PR #780, opened 2023-06-19
feat(android): 🌟 add support for React Native 0.73

ngocle2497:
# Summary
Starting from React Native v0.73 , all libraries will need to be updated with these two one-liners due to Android Gradle Plugin upgrade
[discussions-671](https://github.com/react-native-community/discussions-and-proposals/issues/671)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -58,6 +58,27 @@ if (isNewArchitectureEnabled()) {
     apply plugin: "com.facebook.react"
 }
 
+def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0].toInteger()
+def shouldUseNameSpace = agpVersion >= 7
+def PACKAGE_PROP = "package=\"com.zoontek.rnpermissions\""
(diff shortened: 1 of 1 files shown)


zoontek:
@MasonLe2497 This library currently supports RN 0.63+, will it work for such a wide ranger of versions?

ngocle2497:
I tested on RN 0.71. It's work. I will try with RN 0.63 tomorrow

ngocle2497:
hey @zoontek , i updated my PR, now this library can support RN < 0.71 and RN >= 0.71 both

zoontek:
@MasonLe2497 Thanks, but duplicate of https://github.com/zoontek/react-native-permissions/pull/782
For the moment, I'd rather keep both without overhead. I think stripping the AndroidManifest namespace in the react native CLI, while autolinking could be the best option to keep compatibility for pre 0.71 without an additional warning.

PR #782, opened 2023-06-20
RN-0.73 change Android configs for RN 0.73 compatibility

FelipeSSantos1:
# Summary

Starting from React Native v0.73 , all libraries will need to be updated with these two one-liners due to Android Gradle Plugin upgrade
https://github.com/react-native-community/discussions-and-proposals/issues/671

---

React Native 0.73 will depend on Android Gradle Plugin (AGP) 8.x. This will require all the libraries to specify a namespace in their build.gradle file.

Details
I'd like to share some of the upcoming changes that will happen in 0.73, which is still a bit far, but we'd rather start earlier rather than later.

React Native 0.73 will depend on Android Gradle Plugin (AGP) 8.x, which brings a lot of improvements for Android apps but also a series of notable changes.

Most importantly:

App/Library Developers will now have to install and use [Java 17](https://developer.android.com/build/releases/gradle-plugin#jdk-17-agp)
Library Developers will have to specify a namespace in [their build.gradle file](https://developer.android.com/build/releases/gradle-plugin#namespace-dsl).
Specifically, the last change is a breaking change and will make libraries that are not specifying a namespace incompatible with React Native 0.73 (your project won't build).

Support for namespace was added in AGP 7.3.x, which ships with React Native 0.71. Libraries that published a new version with a namespace declared for 0.71 or 0.72 don't need further update. So we invite library authors to do those changes as soon as possible so by the time 0.73 is out, most of the apps are adapted.

What you need to change
Library authors will have to update their `android/build.gradle` file as follows:

```diff
android {
+   namespace = "com.iterable.reactnative"
    ...
}
```
and remove the package definition from their `AndroidManifest.xml`:

```diff
<?xml version="1.0" encoding="utf-8"?>
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-          package="com.iterable.reactnative">
+          >
...
</manifest>
```
Further reading
[Official Google documentation on namespaces is here](https://developer.android.com/build/configure-app-module#set-namespace).

just build it

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅    |

## Checklist

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -59,6 +59,7 @@ if (isNewArchitectureEnabled()) {
 }
 
 android {
+    namespace = "com.zoontek.rnpermissions"
     // Used to override the NDK path/version on internal CI or by allowing
     // users to customize the NDK path/version from their root project (e.g. for M1 support)
     if (rootProject.hasProperty("ndkPath")) {
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -1,4 +1,4 @@
 <manifest
   xmlns:android="http://schemas.android.com/apk/res/android"
-  >
+  package="com.zoontek.rnpermissions">
 </manifest>
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -1,4 +1,4 @@
 <manifest
   xmlns:android="http://schemas.android.com/apk/res/android"
-  package="com.zoontek.rnpermissions">
+  >
 </manifest>


FelipeSSantos1:
I added back the namespace to AndroidManifest to keep the previous RN compatibility
https://github.com/react-native-community/discussions-and-proposals/issues/671#issuecomment-1597386954

zoontek:
@felipemillhouse Thanks!

osamahafez:
@felipemillhouse build fails on previous RN versions 
```
1: Task failed with an exception.

* Where:
Build file '/Users/osamahafez/Software/ThatekWork/mobile/thatek-mobile/node_modules/react-native-permissions/android/build.gradle' line: 62

* What went wrong:
A problem occurred evaluating project ':react-native-permissions'.
> Could not set unknown property 'namespace' for extension 'android' of type com.android.build.gradle.LibraryExtension.

```
also downgrading to older releases does not fix the issue except 2.0.0 and below

FelipeSSantos1:
I fell it’s not related to the PR @osamahafez

osamahafez:
@felipemillhouse you can see the issue on line 62 where you submitted the PR
here's the commit: https://github.com/zoontek/react-native-permissions/commit/c037cf19c60f55f24bd863ac307f31c886fb249d

Also I submitted an issue here https://github.com/zoontek/react-native-permissions/issues/787 if you like to investigate further on the matter

FelipeSSantos1:
Is that change released in all versions that you seeing failing? I don’t think so. It’s a new change and it shouldn’t impact previous releases @osamahafez 
You can test it checking the node modules folder, deleting the namespace there and trying to build it, likely it’ll fail also

FelipeSSantos1:
If we confirm the root cause is that change, we need to submit another PR like:
```kotlin
// Conditional for compatibility with AGP <4.2.
    if (project.android.hasProperty("namespace")) {
        namespace = "io.sentry.react"
    }
```

osamahafez:
@felipemillhouse I tested it after removing the line `namespace = "com.zoontek.rnpermissions"`
and build was successful

FelipeSSantos1:
> @felipemillhouse I tested it after removing the line `namespace = "com.zoontek.rnpermissions"`
> and build was successful

Could you make a favor to test with this code?
```kotlin
if (project.android.hasProperty("namespace")) {
        namespace = "com.zoontek.rnpermissions"
    }
``` 

I’m far from the computer today and tomorrow, if you could test it would be excellent!
Also few free to send a PR if you are able to, otherwise I’ll send it on Tuesday

PR #786, opened 2023-07-01
Target iOS 12.0

hsjoberg:
# Summary

Hey.
`value` which is available in iOS >= 12.0 is being used, causing compilation to fail on newer XCode versions.

image

This PR bumps the iOS target to version 12.0.

I'm not sure about tvOS.

## Checklist

### DIFF
--- a/RNPermissions.podspec
+++ b/RNPermissions.podspec
@@ -1,46 +1,26 @@
-require 'json'
-package = JSON.parse(File.read('./package.json'))
+require "json"
 
-fabric_enabled = ENV['RCT_NEW_ARCH_ENABLED'] == '1'
+package = JSON.parse(File.read(File.join(__dir__, "package.json")))
 
 Pod::Spec.new do |s|
-  s.name                      = "RNPermissions"
+  s.name         = "RNPermissions"
 
-  s.version                   = package["version"]
-  s.license                   = package["license"]
-  s.summary                   = package["description"]
-  s.authors                   = package["author"]
-  s.homepage                  = package["homepage"]
+  s.version      = package["version"]
+  s.license      = package["license"]
+  s.summary      = package["description"]
+  s.author       = package["author"]
+  s.homepage     = package["homepage"]
 
-  s.ios.deployment_target     = "10.0"
-  s.tvos.deployment_target    = "11.0"
-  s.requires_arc              = true
+  s.requires_arc = true
 
-  s.source                    = { :git => package["repository"]["url"], :tag => s.version }
-  s.source_files              = "ios/*.{h,m,mm}"
+  s.source       = { :git => package["repository"]["url"], :tag => s.version }
+  s.source_files = "ios/*.{h,m,mm}"
 
-  if fabric_enabled
-    folly_compiler_flags      = '-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -Wno-comma -Wno-shorten-64-to-32'
-
-    s.pod_target_xcconfig     = {
-      'HEADER_SEARCH_PATHS' => '"$(PODS_ROOT)/boost" "$(PODS_ROOT)/boost-for-react-native" "$(PODS_ROOT)/RCT-Folly"',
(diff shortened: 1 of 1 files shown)


mikehardy:
That's a compilation error from RCTTypeSafety, which is a react library.
I believe the proposed patch here is invalid then, since it is not related to this library.

Please search first (it shows up in the first results on a web search as a react-native issue already fixed in later versions - https://github.com/facebook/react-native/issues/34106

hsjoberg:
@mikehardy This issue is affecting multiple libs across the ecosystem. For example react-native-svg: https://github.com/software-mansion/react-native-svg/pull/2041

Yes, I'm aware of the GitHub issue about `react_native_pods.rb` -- I am already running react-native 0.72.1 (this is from a fresh installation).

Furthermore, it looks like RCTTypeSafety sets 12.4 as the iOS version.
https://github.com/facebook/react-native/blob/v0.72.1/packages/react-native/Libraries/TypeSafety/RCTTypeSafety.podspec#L26

So I still believe this PR is valid, although I agree it's a bit weird.
You can try for yourself with XCode 14.3.

mikehardy:
This PR is a breaking change that causes this library to no longer support older versions, and is related to code that is not in the library.

It is off-target here, I do not believe it is valid.

Use a Podfile post_install step to modify the target versions if you need it, I think.

mikehardy:
I verified this would drop support for anyone on react-native < 0.69, 0.68 supported iOS11
https://github.com/facebook/react-native/commit/c71e6efbcd2b95faee327d9763d321488120bc5e

This may be something @zoontek agrees with, it may not, but it seems pretty strange for me for a react-native issue to force libraries to adopt a breaking change that cuts off older versions - this is part of what forces people to prematurely landfill devices for no good reason

RNPermissions.podspec
```diff
@@ -12,7 +12,7 @@ Pod::Spec.new do |s|
   s.authors                   = package["author"]
   s.homepage                  = package["homepage"]
 
-  s.ios.deployment_target     = "10.0"
+  s.ios.deployment_target     = "12.0"
```
numandev1: to avoid break changes, I think, it should be like this `s.ios.deployment_target =(fabric_enabled ? "12.0" : "10.0")`

numandev1 (changes requested):

RNPermissions.podspec
hsjoberg: Sure, yeah this seems reasonable. I didn't understand that this only affects new architecture (still not sure why).

hsjoberg:
I've updated the PR as per @numandev1's suggestion.

zoontek:
@hsjoberg Hi, and thanks for the PR! I have a small suggestion: maybe we should migrate to `install_modules_dependencies` too (to fit what I do in my others libs podspec), like this:

```ruby
require "json"

package = JSON.parse(File.read(File.join(__dir__, "package.json")))

Pod::Spec.new do |s|
  s.name         = "RNPermissions"

  s.version      = package["version"]
  s.license      = package["license"]
  s.summary      = package["description"]
  s.author       = package["author"]
  s.homepage     = package["homepage"]

  s.requires_arc = true

  s.source       = { :git => package["repository"]["url"], :tag => s.version }
  s.source_files = "ios/*.{h,m,mm}"

  if ENV['RCT_NEW_ARCH_ENABLED'] == '1' then
    install_modules_dependencies(s)
    s.platforms  = { :ios => "12.4", :tvos => "12.4" }
  else
    s.dependency   "React-Core"
    s.platforms  = { :ios => "10.0", :tvos => "11.0" }
  end
end
```

This solve the issue without any breaking and its IMHO a cleaner way to handle both archs. Could you make the change and copy this instead?
PS: We target 12.4 since that's the version targeted by `react-native` since `v0.69.0`

hsjoberg:
@zoontek sure I'll fix that.

hsjoberg:
I've changed to suggested code and tested building with new arch both activated and inactivated.

zoontek (approved):

zoontek:
Thanks, it's published under 3.8.4: https://github.com/zoontek/react-native-permissions/releases/tag/3.8.4

PR #805, opened 2023-09-12
Silently Fail onRequestPermissionsResult for Invalid Activities

webraptor:
# Summary

react-native-permissions causes crashes to android users in different situations that are hard to replicate, but are clearly visible through crashlogs and analytics, affecting up to 15% of users on different devices and OS versions, with different versions of react-native-permissions in the past 6+ months.

## Crash report
```
com.zoontek.rnpermissions.RNPermissionsModule.onRequestPermissionsResult
java.lang.NullPointerException: Attempt to invoke interface method 'void com.facebook.react.bridge.Callback.invoke(java.lang.Object[])' on a null object reference
```

## Similar reported issues / topics

The issue has been fixed by react-native in a previous release, based on this [PR](https://github.com/facebook/react-native/pull/37047) and it's [changed files](https://github.com/facebook/react-native/pull/37047/files). Current PR only aligns react-native-permissions with the main fix released on react-native.

- https://github.com/facebook/react-native/issues/29056

The impact of the change is minimal.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅     |

## Checklist

- [X] I have tested this on multiple devices and simulators

### DIFF
--- a/android/src/main/java/com/zoontek/rnpermissions/RNPermissionsModule.java
+++ b/android/src/main/java/com/zoontek/rnpermissions/RNPermissionsModule.java
@@ -411,9 +411,13 @@ public class RNPermissionsModule extends NativePermissionsModuleSpec implements 
 
   @Override
   public boolean onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
-    mCallbacks.get(requestCode).invoke(grantResults, getPermissionAwareActivity());
-    mCallbacks.remove(requestCode);
-    return mCallbacks.size() == 0;
+    try {
+      mCallbacks.get(requestCode).invoke(grantResults, getPermissionAwareActivity());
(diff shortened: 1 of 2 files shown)


zoontek:
Thanks, merged! (available in [3.9.1](https://github.com/zoontek/react-native-permissions/releases/tag/3.9.1))

SunnyBalani92:
Hi @zoontek , we have also faced this issue recently but we can't update the library directly to 3.9.1 as our current app is made with 3.8.0. Just wanted to know that if we just add the try catch fix done in 3.10.1(latest version) in our app before releasing. So will that work or it can create any problem?

zoontek:
@SunnyBalani92 Just update, it's easier.

PR #846, opened 2024-02-18
replace deprecated microphone api

zoontek:


ridwansameer:
Hey @zoontek Not sure if this is the best place for this (Lmk if you rather me create an issue), But ever since we (accidentally) upgraded to 4.1.2 we're receiving these errors that look like they are because of this PR

image

Happening on an EAS Build as well as local

ridwansameer:
Not sure if this is an issue for building to iOS 16, it should still be supported right?

zoontek:
Be sure to compile with the latest version of Xcode (that supports iOS 17). It's [a requirement since 4.0.0](https://github.com/zoontek/react-native-permissions/releases/tag/4.0.0), maybe you just didn't use the permissions that require that until now.

ridwansameer:
Yep sorry just to chime in here that it was indeed an Xcode issue. My colleague using xcode 14 couldn't build it and neither could EAS which defaults to Xcode 14. Bumped EAS to use Xcode 15 and it works. Sorry about that! :)

(Anyone seeing this later, just add `"image": "latest"` to your `eas.json` under `ios` for your selected profile


```json
"development-simulator": {
      "developmentClient": true,
      "distribution": "internal",
      "ios": {
        "simulator": true,
        "image": "latest"
      }
},
```

pierroo:
not sure I understand this release, does it mean that version prior to 4.1.2 will crash or have error if they request for microphone permission on ios>17 ?

ghashi:
@pierroo , I think the answer is no.
According to this PR, the code will use AVAudioSession prior to ios 17 and will use AVAudioApplication for ios 17+.
it seems AVAudioSession is not deprecated https://developer.apple.com/documentation/avfaudio/avaudiosession . So the code prior to 4.1.12 will continue to work

ps: An advantage of using AVAudioApplication (instead of AVAudioSession): you can request this permission on the iOS simulator. https://twitter.com/zoontek/status/1759214416376033448

Issue #830, opened 2023-12-18
Apple privacy manifest file 


AntonyARHS:
### Why it is needed?

Hi,

I would like to ask a question regarding the new permission file from Apple: https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api

As we don't know what APIs are used internally in your library, could you confirm whether or not you will be managing the new permission file required by Apple? Is the point already open / track somewhere maybe ? 

Thank you very much,

zoontek:
> As we don't know what APIs are used internally in your library

What do you mean? The code is fully open source, and in the `ios` directory

AntonyARHS:
Yes of course, it was more to say you know better than me all the API used in the lib :D I would like to avoid analysing all the dependency source code one by one to make sure there will be no problem. So it was just in case you were aware of the topic and already had an answer :) 

PS : Sorry, I didn't think to adapt my message on the issue. It was a copy and paste of a request made to another non-open source SDK.

zoontek:
Only `NSUserDefaults` is used by the `FaceID` permission.

yamade:
Hello @zoontek Do you have a plan to add privacy manifest file for `NSUserDefaults` usage soon? 

Ref: https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278401

zoontek:
@yamade It's the app author job (it should be the `CA92.1`)

fabiendem:
Hey @zoontek, sorry, but are you sure about "It's the app author job"?

Reading: https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api

See the highlighted section:
<img width="954" alt="Screenshot 2024-02-22 at 12 42 55" src="https://github.com/zoontek/react-native-permissions/assets/1186282/3f703f8e-7df4-4e98-a2a0-6ca2f4db588f">

> If you use the API in your third-party SDK’s code, then you need to report the API in your third-party SDK’s privacy manifest file. **Your third-party SDK can’t rely on the privacy manifest files for apps that link the third-party SDK**, or those of other third-party SDKs the app links, to report your third-party SDK’s use of required reasons API.

My understanding is that Apple is annoyingly asking all the SDK maintainers to declare their own manifest.
(I am not using the FaceID permission, not pushing here).

Do you think there is a way to dynamically add the declaration based on the `setup_permissions` parameters?

[...]

PR #861, opened 2024-03-24
add apple privacy manifest & usage descriptions for UserDefaults

johnhatvani:
# Summary

#830 

Starting May 1, Apple will enforce 3rd party SDK's to sign & provide privacy manifests for specific API's being accessed by the code. This change adds a privacy manifest and reason for using the data. [Apple Documentation](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api)

I have added a declaration for use of `UserDefualts` via `NSPrivacyAccessedAPICategoryUserDefaults` and from reading the code have stated the reason 
```
CA92.1
Declare this reason to access user defaults to read and write information that is only accessible to the app itself.

This reason does not permit reading information that was written by other apps or the system, or writing information that can be accessed by other apps.
```

Build and run the example app

Upon building the IPA you can inspect the IPA and find `RNPermissions_PrivacyInfo.bundle` and within that find `PrivacyInfo.xcprivacy`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

## Notes
- iOS Only
- No changes required to documentation
- No additions required for the example project

### DIFF
--- a/RNPermissions.podspec
+++ b/RNPermissions.podspec
@@ -17,6 +17,7 @@ Pod::Spec.new do |s|
   s.source       = { :git => package["repository"]["url"], :tag => s.version }
   s.source_files = "ios/*.{h,mm}"
   # s.frameworks = <frameworks>
+  s.resource_bundles = { 'RNPermissions_PrivacyInfo' => 'ios/Resources/PrivacyInfo.xcprivacy' }
 
   if ENV['RCT_NEW_ARCH_ENABLED'] == "1" then
     install_modules_dependencies(s)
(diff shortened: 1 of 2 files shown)


zoontek:
@johnhatvani Thanks! I merge that in a separate branch as it's still need a bit of work (ideally, we want to add the manifest only if the concerned permission is used: FaceID)

Then I will do a release.

johnhatvani:
@zoontek Thanks! I'm not sure that the manifests are conditional like that, unless you have seen something different?

zoontek:
@johnhatvani I updated the setup script to conditionally comment / uncomment the line: https://github.com/zoontek/react-native-permissions/blob/master/scripts/setup.rb

johnhatvani:
nice change! was wondering how you would achieve that 🙏🏼  -- will that code be omitted if the end-user does not use FaceID? - I would assume that the scanner apple have won't look for when a particular API is utilised, only scan the binary for the symbol being present in the code

PR #877, opened 2024-05-27
Compatible with AGP versions below 4.2.0

WzDTj:
# Summary
react-native-permission uses lambda expressions and cannot be compiled in projects with AGP versions lower than 4.2.0.
Therefore, add compileOptions to compile with Java 8.

https://developer.android.com/build/releases/past-releases/agp-4-2-0-release-notes

AGP versions lower than 4.2.0

Compiling with AGP lower than 4.2.0 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -25,6 +25,11 @@ android {
     buildToolsVersion safeExtGet("buildToolsVersion", "33.0.0")
     compileSdkVersion safeExtGet("compileSdkVersion", 33)
 
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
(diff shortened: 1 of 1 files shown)


zoontek:
@WzDTj Hey, this shouldn't occur on currently supported React Native versions, so please update or use an older version. More infos:

- https://github.com/zoontek/react-native-permissions?tab=readme-ov-file#support
- https://github.com/reactwg/react-native-releases/blob/main/docs/support.md

PR #878, opened 2024-06-03
feat(android): add schedule exact alarm permission

seyedmostafahasani:
# Summary
I managed the new permission for scheduling exact alarms in Android 13 and higher.

## Test Plan
I added this permission to the sample app.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅    |

## Checklist

- [X] I have tested this on a device and a simulator
- [X] I added the documentation in `README.md`
- [X] I added a sample use of the API in the example project (`example/App.tsx`)

Issue #887, opened 2024-07-26
Support for Android private / signature permissions


Qwin:
### Why it is needed?

Currently the Permission type does not support private/signature permissions like com.google.android.gms.permission.CAR_VENDOR_EXTENSION, I would like to ask to support those permissions as well or at least just allow freeform as an overload. 

### Possible implementation

requestMultiple([PERMISSIONS.ANDROID.BLUETOOTH_CONNECT, PERMISSIONS.ANDROID.BLUETOOTH_SCAN, "com.google.android.gms.permission.CAR_VENDOR_EXTENSION"])

just a note but the above does not work sadly I get unavailable and never get a prompt when I forcibly type it as Permission, while using AndroidPermissions (default react-native library) does work with the above code for the private permission

**UPDATE**:

looking at the code a bit closer I think the issue is with this method here not accounting for privledged / signature permissions:
<img width="891" alt="image" src="https://github.com/user-attachments/assets/9a3f128d-be9f-46ae-99c6-b6c9b82e85c7">

**UPDATE 2**

So figured out the issue it seems that https://developer.android.com/reference/android/Manifest.permission doesnt have the CAR_VENDOR_EXTENSION permission in it due to it being a system level permission only given out to some companies. Because of that I propose the following solution:

```
requestMultiple(permission, { skipManifestCheck: true }), that way you can skip the check above and avoid it from being unavailable. 
```

**UPDATE 3**

looking at the permission itself based on this post, it seems to be a dangerous permission:
https://stackoverflow.com/questions/32061034/how-do-we-list-all-defined-permissions

zoontek:
@Qwin This is not an issue anymore [in v5.0.0](https://github.com/zoontek/react-native-permissions/pull/890#issuecomment-2346411328), as the `UNAVAILABLE` status has been removed (merged with `BLOCKED`) - like the builtin `PermissionsAndroid` module does with `NEVER_ASK_AGAIN`).

You need to cast the type:

```ts
check("com.google.android.gms.permission.CAR_VENDOR_EXTENSION" as Permission)
```

zoontek:
This is fixed in [v5.0.0](https://github.com/zoontek/react-native-permissions/releases/tag/5.0.0). As it's not exposed, you need to cast it:

```ts
requestMultiple("com.google.android.gms.permission.CAR_VENDOR_EXTENSION" as Permission)
```

PR #890, opened 2024-09-03
v5.0.0

zoontek:
As iOS 18 / Android 15 are around the corner, it's time to publish a new major version, as it will needs the latest Xcode version (v16) to be built. Let's perform some changes to improve this library in the same time.

I'm opening this to discuss about some points. I think it will be for the better, but I would love a community confirmation about them:

- Add limited contacts support (on iOS 18+) ✅
- Drop support for iOS < 13.4 ✅
- Drop support for Android < 6.0 ✅
- ~Remove the `UNAVAILABLE` status, return `BLOCKED` instead (as it cannot be requested). That's one less thing to check, and it allows me to solves [this issue](https://github.com/zoontek/react-native-permissions/issues/887).~
- Migrate the Android code to Kotlin ✅
- ~Make `check` and `checkMultiple` synchronous~
- ~Make `check` and `checkMultiple` return a `boolean` (granted) (as Android will never return `blocked` on check, you have to call request to get the info anyway)~

And, this is extra, but:

- Add hooks. `const [granted, request] = usePermission(…)` might be a thing, let's see how we can make it reactive.
- Allow location permission escalation (closes https://github.com/zoontek/react-native-permissions/pull/716) ✅
- Add `SCHEDULE_EXACT_ALARM` permission (closes https://github.com/zoontek/react-native-permissions/pull/878)


fobos531:
About hooks and reactivity, especially in context of permissions, this github thread at expo has some interesting insight on the API design: https://github.com/expo/expo/pull/31171#discussion_r1731486654

guillempuche:
Hooks!

zoontek:
I published a beta version, try it: `yarn add react-native-permission@next` / `npm i --save react-native-permission@next`

### Beta changelog

- Drop support for React native < 0.73, iOS < 13.4 and Android < 6
- Add `LIMITED` contacts permissions support (requires Xcode 16 beta)
- Rewrite android codebase in Kotlin
- Remove the `UNAVAILABLE` status, `BLOCKED` is now returned in such cases
- Unify all platforms to a unique permission flow
- `check`, `checkMultiple` and `checkNotifications` now return `granted` `boolean` instead of statuses (as it was not truly possible on Android to determine if it was `BLOCKED` without requesting it)
- `requestNotifications` now accept a `rationale` argument

**EDIT:** Just published a second beta that switch back to async `check` / `checkMultiple`, as after some tests it was too slow for some iOS permissions (Face ID…), and was locking the JS thread. Plus, it reduces the amount of breaking changes / add back working windows compat.

android/src/main/java/com/zoontek/rnpermissions/RNPermissionsModuleImpl.kt
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    try {
+      val activity = getPermissionAwareActivity(reactContext)
+
+      callbacks.put(
+        requestCode,
+        Callback { args ->
+          val results = args[0] as IntArray
+          val callbackActivity = args[1] as PermissionAwareActivity
```
TheRogue76, line 215: Is this a safe operation?

zoontek: Yes, it was the same in the  Java codebase / in react native core

zoontek:
Just published beta 2 to fix an issue with bob + the expo plugin (more infos https://github.com/callstack/react-native-builder-bob/pull/598#issuecomment-2350930474)

valeeum:
I don't quite understand what the flow would be to check and then request permissions if the check method returns a boolean.  what if a user provides limited permissions but I want to force granted permissions.  how would i know that permissions are limited and present user to click on a button to be redirected to Linking.openSettings(). 

Right now, i have this block of code.   But if a user is presented this for the first time and they select limited permissions, then once request completes, they will be redirected to settings immediately.  

```
const contactsPermissionResult = await check(PERMISSIONS.IOS.CONTACTS);
if (!contactsPermissionResult) {
  request(PERMISSIONS.IOS.CONTACTS)
    .then(status => {
      if (status === RESULTS.GRANTED) {
        //...dismiss modal
      } else {
        Linking.openSettings();
      }
    })
  }
```

zoontek:
@valeeum The flow is aligned with the Android one.

In your example, `request` resolve with a `PermissionStatus`, so you can verify if it's `GRANTED` or `LIMITED`, and no need to redirect the user to settings in such case.

zoontek:
Published beta 3 with iOS location permission upgrade support.

martinpoulsen:
I have verified the iOS location permission upgrade from "when in use" to "always", it works great 👍 Thank you for adding that 🙏

martinpoulsen:
In my company's use case, we differentiate the UI/UX context for permission requests to a high degree based on both platform and whether it's possible to request a given permission from within the app vs sending the user to settings. With the changed signature for `check` and `checkMultiple`, we loose the information we need for this UI/UX customisation.

I understand the problem with the current `check` signature where the returned `PermissionStatus` is sometimes misleading e.g. in case of Android (due to the native APIs only reveals whether the permission is granted or not) so I see the motivation for the change. However, we would be very sad to lose this information from the iOS permission checks.

I have a suggestion for a different solution - inspired by the signature of `checkNotifications`. Could `check` and `checkMultiple` have a simlar response structure containing both a `granted` boolean and an optional `status` in case the platform APIs support it? I.e.

```
check: (permission: Permission) => Promise<{granted: boolean; status?: PermissionStatus}>
```

zoontek:
@martinpoulsen You are not the only company that reached me about this.

I rolled back to `UNAVAILABLE` (as it's useful in some context, for example to detect if the device has Face ID, etc) and `status`, but managed to solve the calendar permission escalation issue on iOS.

As the Android limitation is not well understood, I also improved the documentation with a unique flow, and a flow implementation example:

Screenshot 2024-10-05 at 19 14 43

I also used GH flavored markdown to add important alerts:

Screenshot 2024-10-05 at 19 18 23

This should be good for the stable release. I published a RC version, will publish the stable one next week.

zoontek:
I published it, will see for other features (hooks, the `SCHEDULE_EXACT_ALARM` permission) later, in a minor version.

rukchatpoomw:
I tried it then I still curious why the  `POST_NOTIFICATIONS` is gone. 
I attached image that display `POST_NOTIFICATIONS` is used. if it's need to remove from react-native-permission so how I solve the case?
Screenshot 2567-10-28 at 11 56 51

arlovip:
> I tried it then I still curious why the `POST_NOTIFICATIONS` is gone. I attached image that display `POST_NOTIFICATIONS` is used. if it's need to remove from react-native-permission so how I solve the case? Screenshot 2567-10-28 at 11 56 51

Facing the same issue. I couldn't find  `POST_NOTIFICATIONS` permission for Android after upgrading to `5.0.0`.

zoontek:
@rukchatpoomw @arlovip Because `POST_NOTIFICATIONS` is not retro-compatible / multi-platform. The preferred way has always been `checkNotifications` / `requestNotifications` that handle both of these.

arlovip:
> @rukchatpoomw @arlovip Because `POST_NOTIFICATIONS` is not retro-compatible / multi-platform. The preferred way has always been `checkNotifications` / `requestNotifications` that handle both of these.

Get it now, thanks a lot!

zoontek:
@arlovip Yeah, I removed it as a lot of people didn't know about `checkNotifications` / `requestNotifications`, just performed a runtime check on the Android version and handle the Android 13+ case, assuming that your app notifications cannot be disabled on Android 12-. But they can.

Issue #911, opened 2024-12-05
iOS - request PERMISSIONS.IOS.CALENDARS does not return the correct permission status


mysport12:
### Before submitting a new issue

- [X] I tested using the latest version of the library, as the bug might be already fixed.
- [X] I tested using a [supported version](https://github.com/reactwg/react-native-releases/blob/main/docs/support.md) of react native.
- [X] I checked for possible duplicate issues, with possible answers.

### Bug summary

Requesting the PERMISSIONS.IOS.CALENDARS and selecting 'Full Access' from the OS prompt returned a status code of 0 from [EKEventStore authorizationStatusForEntityType:EKEntityTypeEvent] in the completion handler resulting in 'denied' on the TS side. It seems as though for iOS17+ and the new method of requesting access, that the BOOL returned in the completion handler needs to be used to determine permission. I have modified the completion handler (screenshot below) and the request now returns the proper permission status based upon what is chosen in the OS prompt.

Additional Notes:
If I kept pressing the button in my app that calls the request function, after two or three times it would eventually return granted, but further requests to check or request the same permission would be inconsistent. After doing a bit of Google research, other examples of this workflow (requesting calendar permissions on iOS) only use a single instance of the EKEventStore instead of a new one each time a request for permission is made. Not sure if that is something worth exploring??

Other relevant env info:

iOS 18.1 physical iPhone 15 Pro
RN 0.76.3 new arch enabled

<img width="799" alt="image" src="https://github.com/user-attachments/assets/6d4cf940-d866-4b80-8957-5c88b8b90dda">

### Library version

5.2.1

### Environment info

```shell
System:
  OS: macOS 15.1.1
  CPU: (12) arm64 Apple M3 Pro
  Memory: 113.33 MB / 36.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 23.3.0
    path: /opt/homebrew/bin/node
  Yarn:
    version: 4.5.0
    path: /opt/homebrew/bin/yarn
  npm:
    version: 10.9.0
    path: /opt/homebrew/bin/npm
  Watchman:
    version: 2024.11.18.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.15.2
    path: /Users/craig/.rvm/gems/ruby-3.3.5/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 24.1
      - iOS 18.1
[...]

PR #929, opened 2025-02-14
Return Limited / Granted on request, instead of always Granted

jlo1:
# Summary

When requesting Contacts permissions on iOS, we'd return `RESULTS.GRANTED`  even when the user provides `LIMITED` permissions.

Instead of returning granted based on boolean return, just return the status directly

## Test Plan
- Reset permissions in app `xcrun simctl privacy booted reset contacts <APP_ID>`
- Allow limited on first go --> should return limited. Allow full access on first go --> should return granted

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

### DIFF
--- a/ios/Contacts/RNPermissionHandlerContacts.mm
+++ b/ios/Contacts/RNPermissionHandlerContacts.mm
@@ -43,7 +43,7 @@
     if (error != nil && error.code != 100) { // error code 100 is permission denied
       reject(error);
     } else {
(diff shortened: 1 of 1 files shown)


zoontek:
Issue here is that relying on `status` is flaky: https://github.com/zoontek/react-native-permissions/issues/911

zoontek:
Fixed in [5.5.3](https://github.com/zoontek/react-native-permissions/releases/tag/5.5.3)

Issue #869, opened 2024-04-18
[Android] 'Allow once' permission option always returns granted as permission status even after killing the app and relaunch it again


DaliaElhefny:
### Before submitting a new issue

- [X] I tested using the latest version of the library, as the bug might be already fixed.
- [X] I tested using a [supported version](https://github.com/reactwg/react-native-releases/blob/main/docs/support.md) of react native.
- [X] I checked for possible duplicate issues, with possible answers.

### Bug summary

I'm requesting location permission in my app `ACCESS_FINE_LOCATION` and faced unexpected behavior
When requesting the location permission the OS alert appears and then I select to allow `Only this time` option

**Expected:** 
- When I kill the app and re-open it again then request the location permission again the OS alert should appear again to ask for permission
- When using `check()` function to get permission status it should return `denied` in this case

**Actual:** 
- When I kill the app and re-open it again then request the location permission again the OS alert does not appear again to ask for the permission
- When using `check()` function to get permission status it always returns `granted` no matter how many times I kill the app and re-open it again to check the permission's status

### Library version

4.1.5

### Environment info

```shell
System:
    OS: macOS 14.4
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 17.24 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.0.0 - ~/.nvm/versions/node/v18.0.0/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.6.0 - ~/.nvm/versions/node/v18.0.0/bin/npm
    Watchman: 2023.03.20.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /Users/my-user/.rbenv/shims/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 23.4, iOS 17.4, macOS 14.4, tvOS 17.4, visionOS 1.1, watchOS 10.4
    Android SDK: Not Found
  IDEs:
    Android Studio: 2023.1 AI-231.9392.1.2311.11076708
    Xcode: 15.3/15E204a - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.15 - /Library/Java/JavaVirtualMachines/jdk-11.0.15.jdk/Contents/Home/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 18.2.0 => 18.2.0 
[...]

PR #971, opened 2025-11-04
Update documentation with one-time permission info for Android

hryhoriiK97:
# Summary

There is no clear or detailed explanation in the official Android documentation about how one-time permissions behave after being granted. In practice, these permissions (e.g., location, camera, or microphone) are not revoked immediately when the app is backgrounded or closed - they are typically revoked automatically within 30–60 seconds after the app is no longer in use.

This PR updates the README to document this behavior in the Android setup section. The goal is to help developers better understand how one-time permissions work on Android 11+ and avoid confusion during testing.

This also addresses the following issue: https://github.com/zoontek/react-native-permissions/issues/869#issuecomment-2066015840

## Test Plan

No testing required.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,19 @@ Add all wanted permissions to your app `android/app/src/main/AndroidManifest.xml
 </manifest>
 ```
 
+> On Android 11+:
+>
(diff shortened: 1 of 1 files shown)


zoontek:
Thanks, merged!

PR #978, opened 2026-03-12
refactor(ios): resolve permission handlers at runtime

zhfwch:
# Summary

This PR updates iOS permission handler resolution in `RNPermissions.mm` to use runtime class lookup via `NSClassFromString` instead of compile-time `__has_include(...)` checks.

In our integration, the current `__has_include(...)`-based approach did not reliably resolve optional permission handlers, which could lead to handlers not being found even when the corresponding implementation was present. `RNPermissions.mm` also needs to explicitly import and branch on each optional handler header, which adds a large amount of repetitive conditional code.

This change resolves handlers at runtime from known library-owned class names instead of relying on compile-time optional header checks.

The intended behavior is unchanged:
- standard permission handlers are resolved dynamically from a class-name list
- `Notifications`, `PhotoLibrary`, and `LocationAccuracy` are also resolved at runtime
- existing "pod is missing" rejections are kept unchanged when a handler class is unavailable

This only affects the iOS native implementation. There are no JS API changes and no Android changes.

## Test Plan

Not run before opening this PR.

- An iOS app integrating `react-native-permissions`
- CocoaPods installed
- A setup that can enable or disable individual permission pods

1. Install pods and build the app with a subset of permission handlers enabled.
2. Verify that regular permission checks and requests still behave as before.
3. Verify that missing optional permission pods still produce the existing rejection behavior.
4. Test the iOS-specific entry points with their corresponding pods enabled:
   - `checkNotifications` / `requestNotifications`
   - `openPhotoPicker`
   - `checkLocationAccuracy` / `requestLocationAccuracy`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |     ✅      |
| Android |     ❌      |

## Checklist

### DIFF
--- a/ios/RNPermissions.mm
+++ b/ios/RNPermissions.mm
@@ -1,68 +1,32 @@
 #import "RNPermissions.h"
 #import <React/RCTLog.h>
 
-#if __has_include("RNPermissionHandlerBluetooth.h")
-#import "RNPermissionHandlerBluetooth.h"
-#endif
-#if __has_include("RNPermissionHandlerCalendars.h")
-#import "RNPermissionHandlerCalendars.h"
-#endif
-#if __has_include("RNPermissionHandlerCamera.h")
-#import "RNPermissionHandlerCamera.h"
-#endif
-#if __has_include("RNPermissionHandlerContacts.h")
-#import "RNPermissionHandlerContacts.h"
-#endif
-#if __has_include("RNPermissionHandlerFaceID.h")
-#import "RNPermissionHandlerFaceID.h"
-#endif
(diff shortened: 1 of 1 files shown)


zoontek:
@zhfwch Hello

This defeat the whole point of splitting the permission handlers: when submitting an app to the app store, Apple performs static analysis and block reviews if you don't request some permissions despite having the code to request them.

zhfwch:
I found the cause — this turned out to be related to ccache, not the PR itself.

After disabling or clearing ccache, I could no longer reproduce the issue.

Leaving a note here in case this helps others. Sorry for the confusion.

Issue #895, opened 2024-09-25
Support for iOS 18 Limited Access?


Lee-sungheon:
### Why it is needed?

Hi @zoontek! Thanks for this awesome and very useful library :)

I was wondering if you were planning to add support for the iOS 18 changes with the limited contact access?

https://developer.apple.com/videos/play/wwdc2024/10121/
Here's a reference to the relevant feature.

Thank you!

zoontek:
@Lee-sungheon Yes

Duplicate https://github.com/zoontek/react-native-permissions/issues/894

Beta https://github.com/zoontek/react-native-permissions/pull/890

Issue #894, opened 2024-09-21
In iOS 18, a new feature has been added to the Contacts app, introducing a restricted permission called CNAuthorizationStatusLimited. If this permission is granted, attempting to access contacts will result in no returned records.


chen834921478:
### Before submitting a new issue

- [X] I tested using the latest version of the library, as the bug might be already fixed.
- [X] I tested using a [supported version](https://github.com/reactwg/react-native-releases/blob/main/docs/support.md) of react native.
- [X] I checked for possible duplicate issues, with possible answers.

### Bug summary

RNPermissionHandlerContacts.mm 
- (void)checkWithResolver:(void (^ _Nonnull)(RNPermissionStatus))resolve
                 rejecter:(void (__unused ^ _Nonnull)(NSError * _Nonnull))reject {
  switch ([CNContactStore authorizationStatusForEntityType:CNEntityTypeContacts]) {
    case CNAuthorizationStatusNotDetermined:
      return resolve(RNPermissionStatusNotDetermined);
    case CNAuthorizationStatusRestricted:
      return resolve(RNPermissionStatusRestricted);
    case CNAuthorizationStatusDenied:
      return resolve(RNPermissionStatusDenied);
    case CNAuthorizationStatusAuthorized:
      return resolve(RNPermissionStatusAuthorized);
      case CNAuthorizationStatusLimited:
          return resolve(RNPermissionStatusLimited);
  }
}
 iOS

### Library version

4.1.5

### Environment info

```shell
xxx
```

### Steps to reproduce

1. …
2. …

### Reproducible sample code

```js
check( PERMISSIONS.IOS.CONTACTS);
```

zoontek:
https://github.com/zoontek/react-native-permissions/pull/890

PR #985, opened 2026-06-10
Add openContactsPicker to edit the limited contacts selection on iOS 18+

pablogdcr:
# Summary

iOS 18 added limited contacts access. This adds `openContactsPicker()` to let the user update their selection when the `Contacts` permission is `limited`, same as `openPhotoPicker()` does for the photo library (related: #894, #895).

Apple only exposes this picker as a SwiftUI modifier (`contactAccessPicker`), so the Contacts handler gets a small Swift file presenting it through a `UIHostingController`. The ObjC handler calls it via `NSClassFromString` + a protocol, which works the same with static libs and `use_frameworks!` without touching the module setup. `setup.rb` now globs `*.swift` and links `ContactsUI`, and the podspec declares `swift_version`.

## Test Plan

Tested in the example app on an iPhone 16 Pro simulator (iOS 18.6), with both the default pod install and `USE_FRAMEWORKS=dynamic`.

An iOS 18+ simulator. `Contacts` is already enabled in the example Podfile.

- Request `CONTACTS` and pick "Select Contacts" → status is `limited`
- Tap the new toolbar button → the contact access picker opens, edit the selection, tap Done → the promise resolves
- With full access, denied or not determined → rejects with "Contacts permission isn't limited"
- On iOS < 18 → rejects with "Only available on iOS 18 or higher"
- On Android → throws the JS "Only supported by iOS 18 and above" error

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

- [ ] I have tested this on a device and a simulator (simulator only)

### DIFF
--- a/android/src/main/java/com/zoontek/rnpermissions/RNPermissionsModuleImpl.kt
+++ b/android/src/main/java/com/zoontek/rnpermissions/RNPermissionsModuleImpl.kt
@@ -303,8 +303,8 @@ object RNPermissionsModuleImpl {
     return activity
   }
 
-  fun openContactsPicker(promise: Promise) {
-    promise.reject("Permissions:openContactsPicker", "openContactsPicker is not supported on Android")
+  fun openContactPicker(promise: Promise) {
+    promise.reject("Permissions:openContactPicker", "openContactPicker is not supported on Android")
   }
 
   fun openPhotoPicker(promise: Promise) {
--- a/android/src/main/java/com/zoontek/rnpermissions/RNPermissionsModuleImpl.kt
+++ b/android/src/main/java/com/zoontek/rnpermissions/RNPermissionsModuleImpl.kt
@@ -303,6 +303,10 @@ object RNPermissionsModuleImpl {
     return activity
   }
 
+  fun openContactsPicker(promise: Promise) {
+    promise.reject("Permissions:openContactsPicker", "openContactsPicker is not supported on Android")
+  }
+
   fun openPhotoPicker(promise: Promise) {
     promise.reject("Permissions:openPhotoPicker", "openPhotoPicker is not supported on Android")
   }
--- a/ios/Contacts/RNPermissionHandlerContacts.mm
+++ b/ios/Contacts/RNPermissionHandlerContacts.mm
@@ -74,7 +74,13 @@
       return reject(@"cannot_open_limited_picker", @"Contact access picker is unavailable", nil);
     }
 
-    [(id<RNPermissionsContactsPickerInterface>)picker presentFrom:RCTPresentedViewController()
+    UIViewController *viewController = RCTPresentedViewController();
+
+    if (viewController == nil) {
+      return reject(@"cannot_open_limited_picker", @"No presented view controller to present from", nil);
+    }
+
+    [(id<RNPermissionsContactsPickerInterface>)picker presentFrom:viewController
                                                        completion:^{
       resolve(@(true));
     }];
--- a/ios/Contacts/RNPermissionHandlerContacts.mm
+++ b/ios/Contacts/RNPermissionHandlerContacts.mm
@@ -1,7 +1,13 @@
 #import "RNPermissionHandlerContacts.h"
+#import <React/RCTUtils.h>
 
 #if !TARGET_OS_TV
 #import <Contacts/Contacts.h>
+
+@protocol RNPermissionsContactsPickerInterface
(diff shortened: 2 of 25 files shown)


ios/Contacts/RNPermissionsContactsPicker.swift
```diff
@@ -0,0 +1,43 @@
...
+  @State private var isPresented = false
+  let onDismiss: () -> Void
+
+  var body: some View {
+    Color.clear
+      .contactAccessPicker(isPresented: $isPresented) { _ in }
+      .onAppear { isPresented = true }
+      .onChange(of: isPresented) { _, visible in
+        if !visible {
+          onDismiss()
```
zoontek: When dismissed, the pick close immediately because it crashes:

```
[UISceneHosting-com.zoontek.rnpermissionsexample:UIHostedScene-com.apple.ContactsUI.LimitedLibraryPickerViewService-EA2E3708-C7FA-4401-817C-8B0F1E90C1D0] Update failed: <NSError: 0x116f6e580; domain: FBSceneErrorDomain; code: 1 ("operation-failed"); "Scene update failed."> {
    NSUnderlyingError = <NSError: 0x116e17870; domain: FBWorkspaceScene; code: 1 ("send-message-failed"); "Scene update failed. Client process exited."> {
        NSUnderlyingError = <NSError: 0x116e17960; domain: FBProcessExit; code: 1 ("voluntary"); "The process exited voluntarily.">;
    };
```

It's visible, as there's no dismiss animation. To fix it, wrap it:

```swift
if !visible {
  // Defer to the next runloop so UIKit finishes dismissing the
  // (out-of-process) picker before we tear down its host controller.
  DispatchQueue.main.async {
    onDismiss()
  }
}
```

ios/Contacts/RNPermissionHandlerContacts.mm
```diff
@@ -52,4 +58,30 @@ - (void)requestWithResolver:(void (^ _Nonnull)(RNPermissionStatus))resolve
...
+      return reject(@"cannot_open_limited_picker", @"Contacts permission isn't limited", nil);
+    }
+
+    id picker = NSClassFromString(@"RNPermissionsContactsPicker");
+
+    if (picker == nil) {
+      return reject(@"cannot_open_limited_picker", @"Contact access picker is unavailable", nil);
+    }
+
+    [(id<RNPermissionsContactsPickerInterface>)picker presentFrom:RCTPresentedViewController()
```
zoontek: `RCTPresentedViewController()` can return `nil`. Add a check:

```objc
  UIViewController *viewController = RCTPresentedViewController();

  if (viewController == nil) {
    return reject(@"cannot_open_limited_picker", @"No presented view controller to present from", nil);
  }

  [(id<RNPermissionsContactsPickerInterface>)picker presentFrom:viewController
                                                     completion:^{
  // …
```

ios/Contacts/RNPermissionsContactsPicker.swift
```diff
@@ -0,0 +1,43 @@
...
+}
+
+@available(iOS 18.0, *)
+@objc(RNPermissionsContactsPicker)
+public class RNPermissionsContactsPicker: NSObject {
+  @objc public static func present(
+    from viewController: UIViewController,
+    completion: @escaping () -> Void
+  ) {
+    var host: UIHostingController<ContactAccessPickerView>?
```
zoontek: The `onDismiss` closure strongly captures host, which points back to `controller`, creating a retain cycle that leaks a `UIHostingController` per call; fix this with:

```swift
weak var host: UIHostingController<ContactAccessPickerView>?
```

android/src/main/java/com/zoontek/rnpermissions/RNPermissionsModuleImpl.kt
```diff
@@ -303,6 +303,10 @@ object RNPermissionsModuleImpl {
     return activity
   }
 
+  fun openContactsPicker(promise: Promise) {
```
zoontek: Replace `openContactsPicker` with `openContactPicker` (singular) everywhere. It matches `openPhotoPicker`, but also the native iOS naming: [CNContactPickerViewController](https://developer.apple.com/documentation/contactsui/cncontactpickerviewcontroller) (singular)

zoontek (changes requested):

zoontek:
LGTM, thanks @pablogdcr !

PR #986, opened 2026-07-22
Fix stale iOS calendar/reminders status by sharing a single EKEventStore

flochtililoch:
# Summary

On iOS 17+ **with the New Architecture enabled**, `request(PERMISSIONS.IOS.CALENDARS)` resolves `granted`, but an immediately-following `check(PERMISSIONS.IOS.CALENDARS)` returns `denied`. The same class of bug affects `REMINDERS` and `CALENDARS_WRITE_ONLY`.

This started reproducing consistently after turning on the New Architecture. It's not strictly New-Arch-specific (see root cause below), but the TurboModule call path is fast enough that `check()` runs before EventKit's authorization-status cache settles, so the race is effectively always lost. On the old bridge the extra serialization usually hid it.

**Root cause:** the `Calendars`, `CalendarsWriteOnly` and `Reminders` handlers allocate a brand-new `EKEventStore` on every `request()`. Churning instances exhausts the connections to the system `calaccessd` daemon (`"Client tried to open too many connections to calaccessd. Refusing to open another."`). Once that happens, the static `+[EKEventStore authorizationStatusForEntityType:]` returns a stale value — typically `EKAuthorizationStatusNotDetermined` — for a while after access was actually granted. `currentStatus` maps `NotDetermined` → `denied`, so `request()` reports `granted` (from the completion handler's `granted == YES`) while a separate `check()` reports `denied`.

This matches the widely-reported EventKit behavior and Apple's recommended workaround (a single shared `EKEventStore`):
- https://developer.apple.com/forums/thread/737536 — *"I got full access permission, granted is YES. but the status is NotDetermined"*, *"it caused me [...] Client tried to open too many connections to calaccessd"*, resolved by using a single shared `EventStoreManager.shared` instance.
- Related: #911

**Solution:** reuse a single, long-lived `EKEventStore` (via `dispatch_once`) in the three EventKit handlers, keeping the daemon connection healthy so the authorization status stays consistent between `request()` and `check()`. No public API change.

**Impact:** iOS only — `Calendars`, `CalendarsWriteOnly`, and `Reminders` permission handlers.

- A physical iOS device or simulator running iOS 17 or later.
- The New Architecture enabled (this is where the bug reproduces consistently).
- The `example` app with `Calendars` permission set up (already the case), and calendar access in a "not determined" state (fresh install, or reset via Settings → Privacy & Security → Calendars, or Simulator → Device → Erase All Content and Settings).

1. Run the `example` app on iOS 17+ with the New Architecture enabled.
2. Tap the new **calendar-alert** action in the top app bar. It runs `request(PERMISSIONS.IOS.CALENDARS)` and, as soon as it resolves, `check(PERMISSIONS.IOS.CALENDARS)`, then shows both results in the snackbar.
3. In the system prompt, tap **Allow Full Access**.

- **Before this change:** snackbar shows `{ "requested": "granted", "checked": "denied" }`.
- **After this change:** snackbar shows `{ "requested": "granted", "checked": "granted" }`.

Equivalent code:

    import {check, request, PERMISSIONS} from 'react-native-permissions';

    const requested = await request(PERMISSIONS.IOS.CALENDARS); // tap "Allow Full Access"
    const checked = await check(PERMISSIONS.IOS.CALENDARS);
    console.log({requested, checked});

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |     ✅      |
| Android |     ❌      |

## Checklist


zoontek:
Fixed in [5.6.1](https://github.com/zoontek/react-native-permissions/releases/tag/5.6.1)

